PR #8, opened 2023-09-12
sonarcloud bug complain fix

jpikora-iov:


tests/04_00_02-fedchange.js
```diff
@@ -281,14 +281,14 @@ describe('RSK Federation change', function() {
       await checkFederationAddedNewKeys();
       var addResult = await rskClientOldFed.rsk.bridge.methods.addFederatorPublicKeyMultikey('0xaabb', '0xccdd', '0xeeff').call({ from: getRandomFedChangeAddress() });
```
marcos-iov, line 282:
```suggestion
      const addResultFirstCall = await rskClientOldFed.rsk.bridge.methods.addFederatorPublicKeyMultikey('0xaabb', '0xccdd', '0xeeff').call({ from: getRandomFedChangeAddress() });
      expect(Number(addResult)).to.equal(1); // First call successful
```
Sonar is complaining because you declare this variable here but never use it, then declare it again as const.

I think the point of the test is to make the call twice, and check that the second time it fails. So we could also check the result of the first call

tests/04_00_02-fedchange.js
```diff
@@ -281,14 +281,14 @@ describe('RSK Federation change', function() {
...
       expect(Number(addResult)).to.equal(-10); // not a public key
-      
-      await rskClientOldFed.rsk.bridge.methods.addFederatorPublicKeyMultikey(
+
+      const addResult = await rskClientOldFed.rsk.bridge.methods.addFederatorPublicKeyMultikey(
```
marcos-iov, line 285:
```suggestion
      const addResultSecondCall = await rskClientOldFed.rsk.bridge.methods.addFederatorPublicKeyMultikey(
```

jpikora-iov:
pipeline: run

tests/04_00_02-fedchange.js
jpikora-iov: ohh i fixed it before i read this comment, hope my naming is ok for you :)

jpikora-iov: hmm i'm not sure i understand, the assert was always there...
`expect(Number(addKeysResult)).to.equal(-10); // not a public key`

jeremy-then (dismissed):
LGTM.

marcos-iov (approved):

kelvinator07 (approved):

PR #12, opened 2023-09-27
disable whitelisting

jpikora-iov:


tests/02_00_02-2wp_segwit_compatible.js
```diff
@@ -3,17 +3,19 @@ const { getDerivedRSKAddressInformation } = require('@rsksmart/btc-rsk-derivatio
...
 let rskTxHelpers;
 let rskTxHelper;
 let btcTxHelper;
 
-const fulfillRequirementsToRunAsSingleTestFile = async () => {
+const fulfillRequirementsToRunAsSingleTestFile = async (rskTxHelper) => {
     const latestForkName = rskUtils.getLatestForkName()
+    console.log(`latestForkName: ${JSON.stringify(latestForkName)}`)
     await rskUtils.activateFork(latestForkName);
+    disableWhitelisting(rskTxHelper)
```
marcos-iov, line 18:
```suggestion
    await disableWhitelisting(rskTxHelper)
```

lib/2wp-utils.js
```diff
@@ -238,6 +240,23 @@ const donateToBridge = async (rskTxHelper, btcTxHelper, donatingBtcAddressInform
...
+    expect(addr.slice(2)).to.equal(WHITELIST_CHANGE_ADDR);
+    await rskTxHelper.unlockAccount(addr);
+
+    await sendTxWithCheck(
+        rskTxHelper,
+        bridge.methods.setLockWhitelistDisableBlockDelay(1),
+        WHITELIST_CHANGE_ADDR,
+        (disableResult) =>  expect(Number(disableResult)).to.equal(1)
+    );
+    rskTxHelper.mine(1);
```
marcos-iov, line 257: `sendTxWithCheck` already mines 1 RSK block after sending the transaction. If I'm not mistaken you should be mining the amount of Bitcoin blocks that you passed as a parameter to `setLockWhitelistDisableBlockDelay` (in this case 1).

By the way, I'd put that value as a constant inside this method so it's clear what it means

lib/2wp-utils.js
```diff
@@ -238,6 +240,23 @@ const donateToBridge = async (rskTxHelper, btcTxHelper, donatingBtcAddressInform
     return peginBtcTxHash;
 };
 
+const disableWhitelisting = async (rskTxHelper) => {
+    const latestActiveForkName = await getLatestActiveForkName();
+    const bridge = getBridge(rskTxHelper.getClient(), latestActiveForkName);
+
+    const addr = await rskTxHelper.importAccount(WHITELIST_CHANGE_PK)
```
jeremy-then, line 247:
```suggestion
    const whitelistChangeAddress = await rskTxHelper.importAccount(WHITELIST_CHANGE_PK)
```

lib/2wp-utils.js
```diff
@@ -15,6 +15,8 @@ const ADDRESS_TYPES_CODES = {
 const BTC_TO_RSK_MINIMUM_CONFIRMATIONS = 3;
 const TO_BRIDGE_GAS_PRICE = 2;
 const BRIDGE_ADDRESS = '0x0000000000000000000000000000000001000006';
+const WHITELIST_CHANGE_PK = '3890187a3071327cee08467ba1b44ed4c13adb2da0d5ffcc0563c371fa88259c';
+const WHITELIST_CHANGE_ADDR = '87d2a0f33744929da08b65fd62b627ea52b25f8e';
```
jeremy-then, line 19: These constants are already exported from the `whitelisting.js` file. Please, import them and use them here. No need to duplicate them.
Also check that they are redeclared again in `01_03_51-2wp_release_event.js` and `01_03_52-disable_lock_whitelist.js`. Delete them from there and import them on those files from `whitelisting.js`.

marcos-iov: Should the method in it's entirety be moved to whitelisting.js? Or should it stay here? What do you think?

lib/2wp-utils.js
jpikora-iov: ok. 
so the `sendTxWithCheck` will always mine 1 block. does it make sense to extend `sendTxWithCheck` with  parameter `blocksToMine` to be able to change required blocks to mine? with value `1` as default.

marcos-iov: `sendTxWithCheck` will mine 1 **RSK** block by default. But the parameter received by `setLockWhitelistDisableBlockDelay` is the amount of **Bitcoin** blocks you need to mine for the whitelist disabling to have effect. Does that make sense?

jpikora-iov: yes, i understand.
what i'm asking is, will i ever need to mine more then 1 block for whitelist disabling?
something like
```
    await sendTxWithCheck(
        rskTxHelper,
        bridge.methods.setLockWhitelistDisableBlockDelay(blocksToMine),
        WHITELIST_CHANGE_ADDR,
        (disableResult) =>  expect(Number(disableResult)).to.equal(1)
    );
    if (blocksToMine > 1 ) {
        rskTxHelper.mine(blocksToMine - 1);
    }


or add extra parameter to `sendTxWithCheck`...

marcos-iov: `blocksToMine` parameter is the amount of Bitcoin blocks to mine to have the whitelist disabled, it can be more than 1 yes. But `sendTxWithCheck` will mine Rootstock blocks

marcos-iov (approved):

jeremy-then (approved):
LGTM.

PR #18, opened 2023-11-09
Redeem script tests refactor

ed-iov:
This PR introduces 2 new tests that is repeating the test for 01_05_57-post_hop_active_powpeg_redeem_script.js into 2 new files
03_02_01-last_fork_active_powpeg_redeem_script.js - validates that in the last fork activation the redeem script is correct.
05_02_01-last_fork_active_powpeg_redeem_script.js - validates that in the last fork activation and a federation change the redeem script is correct for the new federation.

### DIFF
--- a/tests/03_02_01-last_fork_active_powpeg_redeem_script.js
+++ b/tests/03_02_01-last_fork_active_powpeg_redeem_script.js
@@ -7,6 +7,7 @@ const redeemScriptParser = require('@rsksmart/powpeg-redeemscript-parser');
 const CustomError = require('../lib/CustomError');
 const removePrefix0x = require("../lib/utils").removePrefix0x;
 const { GENESIS_FEDERATION_ADDRESS, GENESIS_FEDERATION_REDEEM_SCRIPT } = require('../lib/constants');
+let rskClient;
 
 describe('Calling getActivePowpegRedeemScript method after last fork before fedchange', function() {
 
--- a/tests/03_02_01-last_fork_active_powpeg_redeem_script.js
+++ b/tests/03_02_01-last_fork_active_powpeg_redeem_script.js
@@ -1,32 +1,60 @@
 const chai = require('chai');
 chai.use(require('chai-as-promised'));
 const expect = chai.expect;
-
-const rsk = require('peglib').rsk;
 const redeemScriptParser = require('@rsksmart/powpeg-redeemscript-parser');
+const {getRskTransactionHelpers} = require('../lib/rsk-tx-helper-provider');
+const {getBridge, getLatestActiveForkName} = require('../lib/precompiled-abi-forks-util');
 const CustomError = require('../lib/CustomError');
-const removePrefix0x = require("../lib/utils").removePrefix0x;
-const { GENESIS_FEDERATION_ADDRESS, GENESIS_FEDERATION_REDEEM_SCRIPT } = require('../lib/constants');
-let rskClient;
+const {disableWhitelisting} = require('../lib/2wp-utils');
+const removePrefix0x = require('../lib/utils').removePrefix0x;
+const {getBtcClient} = require('../lib/btc-client-provider');
+const rskUtils = require('../lib/rsk-utils');
+const {
+  GENESIS_FEDERATION_ADDRESS,
+  GENESIS_FEDERATION_REDEEM_SCRIPT,
+} = require('../lib/constants');
+
+const fulfillRequirementsToRunAsSingleTestFile = async (rskTxHelper, btcTxHelper) => {
+  const latestForkName = rskUtils.getLatestForkName();
+  await rskUtils.activateFork(latestForkName);
+  await disableWhitelisting(rskTxHelper, btcTxHelper);
+};
 
 describe('Calling getActivePowpegRedeemScript method after last fork before fedchange', function() {
+  let rskTxHelpers;
+  let rskTxHelper;
+  let bridge;
 
-    before(() => {
-      rskClient = rsk.getClient(Runners.hosts.federate.host);
-    });
-  
-    it('should return the active powpeg redeem script', async () => {
-      try{
-        const activePowpegRedeemScript = await rskClient.rsk.bridge.methods.getActivePowpegRedeemScript().call();
-        const activeFederationAddressFromBridge = await rskClient.rsk.bridge.methods.getFederationAddress().call();
-        const addressFromRedeemScript = redeemScriptParser.getAddressFromRedeemScript(
-          'REGTEST', Buffer.from(removePrefix0x(activePowpegRedeemScript), 'hex')
+  before(async () => {
+    const btcTxHelper = getBtcClient();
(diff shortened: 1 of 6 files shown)


jeremy-then (commented):
Get rid of the `peglib`. We have better options instead of it. As a reference, check the `2wp.js` file.

tests/03_02_01-last_fork_active_powpeg_redeem_script.js
```diff
@@ -0,0 +1,60 @@
...
+const rskUtils = require('../lib/rsk-utils');
+const {
+  GENESIS_FEDERATION_ADDRESS,
+  GENESIS_FEDERATION_REDEEM_SCRIPT,
+} = require('../lib/constants');
+
+const fulfillRequirementsToRunAsSingleTestFile = async (rskTxHelper, btcTxHelper) => {
+  const latestForkName = rskUtils.getLatestForkName();
+  await rskUtils.activateFork(latestForkName);
+  await disableWhitelisting(rskTxHelper, btcTxHelper);
```
jeremy-then, line 20: I don't think it is necessary to disable the whitelisting in this test. There's no pegin being made.
You could simplify this function to be:

```js
const fulfillRequirementsToRunAsSingleTestFile = async () => {
  const latestForkName = rskUtils.getLatestForkName();
  await rskUtils.activateFork(latestForkName);
};
```
And remove the `const {disableWhitelisting} = require('../lib/2wp-utils');` line above.

And call it like:

```js
if (process.env.RUNNING_SINGLE_TEST_FILE) {
      await fulfillRequirementsToRunAsSingleTestFile();
    }
```

And the ` const btcTxHelper = getBtcClient();` line would not be necessary. 
And the `const {getBtcClient} = require('../lib/btc-client-provider');` line could be removed as well.

tests/05_02_01-last_fork_active_powpeg_redeem_script.js
```diff
@@ -0,0 +1,87 @@
...
+const bitcoin = peglib.bitcoin;
+const {compareFederateKeys} = require('../lib/federation-utils');
+const {getRskTransactionHelpers} = require('../lib/rsk-tx-helper-provider');
+const redeemScriptParser = require('@rsksmart/powpeg-redeemscript-parser');
+const CustomError = require('../lib/CustomError');
+const removePrefix0x = require('../lib/utils').removePrefix0x;
+const {getBridge, getLatestActiveForkName} = require('../lib/precompiled-abi-forks-util');
+
+// in order to run this as a single test file, it requires a federation change so follow the following command
+// npm run run-single-test-file 04_00_02-fedchange.js,05_02_01-last_fork_active_powpeg_redeem_script.js
```
jeremy-then, line 14: Since we are not using the `fulfillRequirementsToRunAsSingleTestFile` function here, I don't think this comment applies. You can remove it.

tests/05_02_01-last_fork_active_powpeg_redeem_script.js
```diff
@@ -0,0 +1,87 @@
+const chai = require('chai');
+chai.use(require('chai-as-promised'));
+const expect = chai.expect;
+const peglib = require('peglib');
+const bitcoin = peglib.bitcoin;
```
jeremy-then, line 5: I see the `peglib` is still being used here. I would rather add a `publicKeyToCompressed` utility function in `utils.js` file or something like that, so you can get rid of the peglib from here.

jeremy-then (commented):
Left a couple of comments.

tests/05_02_01-last_fork_active_powpeg_redeem_script.js
```diff
@@ -0,0 +1,87 @@
...
+          .call();
+      const activeFederationAddressFromBridge = await bridge.methods.getFederationAddress().call();
+      const addressFromRedeemScript = redeemScriptParser.getAddressFromRedeemScript(
+          'REGTEST',
+          Buffer.from(removePrefix0x(activePowpegRedeemScript), 'hex'),
+      );
+      const newFederationPublicKeys = Runners.hosts.federates
+          .filter((federate, index) => index >= INITIAL_FEDERATION_SIZE)
+          .map((federate) => ({
+            [KEY_TYPE_BTC]: bitcoin.keys.publicKeyToCompressed(
```
jeremy-then, line 49: Actually, this is not needed at all. The federators public keys are already compressed. So you can simply do something like this:

```js
return {
              [KEY_TYPE_BTC]: federate.publicKeys[KEY_TYPE_BTC],
              [KEY_TYPE_RSK]: federate.publicKeys[KEY_TYPE_RSK],
              [KEY_TYPE_MST]: federate.publicKeys[KEY_TYPE_MST],
            };
```

tests/05_02_01-last_fork_active_powpeg_redeem_script.js
ed-iov: Actually it does, since the fulfillRequirementsToRunAsSingleTestFile is done in 04_00_02-fedchange.js that not only set the fork activation but does the fed change.
The comment is to run the file as single, but the other as precondition.

tests/05_02_01-last_fork_active_powpeg_redeem_script.js
ed-iov: I tried this approach and there was a difference in the results, I changed it to put in in the utils.js as you suggested above.

tests/05_02_01-last_fork_active_powpeg_redeem_script.js
ed-iov: removing the publicKeyToCompressed as you mentioned is not required for this test, thank you

jeremy-then (dismissed):
LGTM.

tests/05_02_01-last_fork_active_powpeg_redeem_script.js
```diff
@@ -0,0 +1,89 @@
...
+      const expectedNewFederationAddress =
+        redeemScriptParser.getAddressFromRedeemScript(
+            'REGTEST',
+            p2shErpFedRedeemScript,
+        );
+
+      expect(activePowpegRedeemScript)
+          .to.eq('0x' + p2shErpFedRedeemScript.toString('hex'),
+          );
+
```
nathanieliov, line 76: expect(removePrefix0x(activePowpegRedeemScript)).to.eq(p2shErpFedRedeemScript.toString('hex'));

nathanieliov (commented):
Well done Edgardo! LGTM. Just a suggestion.

tests/05_02_01-last_fork_active_powpeg_redeem_script.js
ed-iov: done

jeremy-then (approved):
LGTM.

nathanieliov (approved):
LGTM

ed-plutotv:
pipeline:run

PR #19, opened 2023-11-30
Adds tbd600 fork

jeremy-then:
*rskj:bridge-btc-tx-index-integration*
*fed:bridge-btc-tx-index-integration*

# Changes:

## Arrowhead related changes:

* We are activating `Arrowhead` after the regression tests and before the integration tests. A new file was created for this `tests/01_07_50-activate-tbd600_fork.js`.

* The main changes on this pr have to do with the new Arrowhead fork. Basically, we needed to include this new fork in the RITs with an activation height so this config could be passed to a fed node while testing, so the new `Arrowhead` logic kicks in once the Arrowhead activation block is reached. The new `arrowhead600` fork config was added to the `test.js` file, with an activation height of `1600` for testing.

* Now, there is no `lockingCap` or `whitelisting`. So, integration tests using the `lockingCap` on this pr had to be moved to be regression tests in a pre `Arrowhead` fork, specifically `Fingerroot`. The `lockingCap` tests where removed from `tests/02_00_05-2wp_version1.js` test file and moved into a new `tests/01_06_53-2wp_version1_exceeding-lockingCap.js` test file.

* Some changes where done to the `tests/04_00_02-fedchange.js` file, because this is an integration test file that is supposed to be running with the latest fork. Since now the latest fork is `Arrowhead`, and the whitelisting does not exist anymore, one of the tests calling the whitelisting was failing because the whitelisting call was being ignored. Then, we proceeded to remove all the whitelisting being done on this file since they are not needed anymore.

* A new `precompiled-arrowhead600` abi is being used here. There is no change on the abi for the `Arrowhead` fork, but it would be confusing to be using the `fingerroot500` abi with the new `Arrowhead` logic so we decided to create a new release for a `precompiled-arrowhead600` abi and use this instead.

* The new fork `Arrowhead600` abi was added to the `lib/precompiled-abi-forks-util.js` file.

## Other changes:

* The `singleTestFileRunner.js` file was updated to include the `forkName`. This is to be used when we are running a test file independenty and we would like to specify what fork should that test file use. In most of the test files we know what fork should be active, hence we are able to hard code it. But on some other test files, like the `2wp.js` file, we are not able to hard code the fork to use because this file is called multiple times when every time it is called a different fork is to be active. This change does not affect the test executions. This `singleTestFileRunner.js` is only used during development in a local environment.  So, that leads to the next change:

* The `lib/tests/2wp.js` file was changed to include a `fulfillRequirementsToRunAsSingleTestFile` function. Nothing else was changed. This change is going to be used during development, locally, when we want to run this test file in isolation. With this change, this test should continue to work as usual when al the tests are running.

The `README.md` file was updated to include the new steps on how to specify a fork name while running a test file in isolation using the `singleTestFileRunner.js` file and the `fulfillRequirementsToRunAsSingleTestFile` function.

The changes on the `tests/01_06_52-post_fingerroot500-2wp_with_locking_cap.js` test file where refactors, replacing the `var` keyword with `let` and `const`. This is because Sonar was complaining about it.


tests/02_00_05-2wp_version1.js
```diff
@@ -344,7 +154,7 @@ describe('Lock funds using peg-in protocol version 1', () => {
 
         expect(rejectedPeginEvent).to.not.be.null;
         expect(rejectedPeginEvent.arguments.btcTxHash).to.equal(peginBtcTxHashWith0xPrefix);
-        expect(rejectedPeginEvent.arguments.reason).to.equal(PEGIN_V1_INVALID_PAYLOAD_REASON);
+        expect(rejectedPeginEvent.arguments.reason).to.equal(LEGACY_PEGIN_MULTISIG_SENDER_REASON);
```
marcos-iov, line 157: Isn't this what Natha fixed? That rejected reason doesn't seem right

tests/02_00_05-2wp_version1.js
```diff
@@ -474,7 +284,7 @@ describe('Lock funds using peg-in protocol version 1', () => {
...
```
marcos-iov, line 287: same

lib/precompiled-abi-forks-util.js
```diff
@@ -13,6 +13,7 @@ const IRIS = 'iris';
 const HOP = 'hop';
 const HOP401 = 'HOP401';
 const FINGERROOT = 'fingerroot';
+const TBD600 = 'tbd600';
```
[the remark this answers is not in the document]
jeremy-then: Partially updated! Until the name is replaced in the rskj and powpeg projects.

test.js
```diff
@@ -86,7 +86,8 @@ global.Runners = {
       iris300: createForkObject('iris', 700),
       hop400: createForkObject('hop', 1000),
       hop401: createForkObject('hop401', 1010),
-      fingerroot500: createForkObject('fingerroot', 1350)
+      fingerroot500: createForkObject('fingerroot', 1350),
+      tbd600: createForkObject('tbd600', 2000)
```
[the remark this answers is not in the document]
jeremy-then: Updated! The rsk blockchain height right before this fork is activated is about 488. I think it would be enough with 1600 for the activation of this fork.

tests/04_00_02-fedchange.js
```diff
@@ -682,7 +682,6 @@ describe('RSK Federation change', function() {
         { address: retiringFederationAddress, amount: bitcoin.btcToSatoshis(17) },
       ];
       
-      await whitelistingAssertionsTestLegacy.assertAddOneOffWhitelistAddress(addresses.btc, outputs.reduce((c, v) => c + v.amount, 0))();
```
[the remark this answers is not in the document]
jeremy-then: Updated!

tests/02_00_05-2wp_version1.js
jeremy-then: Updated!

tests/02_00_05-2wp_version1.js
```diff
@@ -420,7 +230,7 @@ describe('Lock funds using peg-in protocol version 1', () => {
...
```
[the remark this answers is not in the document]
jeremy-then: Updated!

tests/02_00_05-2wp_version1.js
jeremy-then: Updated!

jeremy-then:
rskj:add-new-pegin-tests
fed:bridge-btc-tx-index-integration

tests/02_00_05-2wp_version1.js
```diff
@@ -106,188 +98,6 @@ describe('Lock funds using peg-in protocol version 1', () => {
         expect(Number(finalDestinationAddressBalance)).to.equal(Number(btcEthUnitConverter.btcToWeis(AMOUNT_TO_LOCK_IN_BTC)));
     });
 
-    it('should refund to indicated address when locking cap exceeded using bech32 sender', async () => {
```
nathanieliov, line 109: Shouldn't we test that the locking doesn't exist anymore?

nathanieliov (commented):
Well done. Just a couple of comments.

tests/02_00_05-2wp_version1.js
jeremy-then: Why test something that doesn't exist?

lib/precompiled-abi-forks-util.js
```diff
@@ -13,6 +13,7 @@ const IRIS = 'iris';
 const HOP = 'hop';
 const HOP401 = 'HOP401';
 const FINGERROOT = 'fingerroot';
+const ARROWHEAD = 'tbd600';
```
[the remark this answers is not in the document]
jeremy-then: Not possible yet. This value is sent as part of the configuration for the nodes. This value is not yet replaced in the rskj project, if I replace it here, it will not match and from the rskj perspective, it will not get activated or will take a default value set in rskj if present.

lib/precompiled-abi-forks-util.js
```diff
@@ -36,14 +38,15 @@ const cache = new Map();
  */
 const getBridge = (rskClient, forkName = '') => {
 
-    const { orchid, wasabi100, papyrus200, iris300, hop400, hop401, fingerroot500 } = Runners.common.forks;
+    const { orchid, wasabi100, papyrus200, iris300, hop400, hop401, fingerroot500, tbd600 } = Runners.common.forks;
```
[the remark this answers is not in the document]
jeremy-then: Explained above.

tests/01_06_53-2wp_version1_exceeding-lockingCap.js
```diff
@@ -0,0 +1,196 @@
...
+        // Create legacy type address to use as refund
+        const refundAddressInformation = await btcTxHelper.generateBtcAddress('legacy');
+        const initialRefundAddressBalance = await btcTxHelper.getAddressBalance(refundAddressInformation.address);
+
+        // Create peg-in data
+        const data = [];
+        data.push(createPeginV1TxData(rskDestinationAddress, refundAddressInformation.address));
+
+        // Execute peg-in
+        const AMOUNT_TO_LOCK_EXCEEDING_LOCKIN_CAP = AMOUNT_TO_LOCK_IN_BTC + lockingCapInBtc;
```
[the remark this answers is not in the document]
jeremy-then: Updated!

jeremy-then:
rskj:add-new-pegin-tests
fed:bridge-btc-tx-index-integration

nathanieliov (approved):
LGTM. Just create a task to rename `tbd600` references later.

lib/precompiled-abi-forks-util.js
```diff
@@ -36,13 +39,15 @@ const cache = new Map();
...
-    const { orchid, wasabi100, papyrus200, iris300, hop400, hop401, fingerroot500 } = Runners.common.forks;
+    const { orchid, wasabi100, papyrus200, iris300, hop400, hop401, fingerroot500, arrowhead600 } = Runners.common.forks;
     if(cache.has(forkName)) {
         return cache.get(forkName);
     }
 
     let bridge;
     switch(forkName) {
+        case arrowhead600.name:
+            bridge = precompiledArrowhead600.bridge.build(rskClient);
```
marcos-iov, line 50: You are missing the `break` statement here.

Check also the default case, it's set for fingerroot

nathanieliov: Fixed

marcos-iov: What about the default case?

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

PR #29, opened 2024-01-10
Fixes waitForBlock logic to not stop searching if the blockchain is a…

jeremy-then:
Fixes waitForBlock logic to not stop searching if the blockchain is at least advancing. Unless it gets stuck in a block and the `maxAttempts` gets consumed waiting for the blockchain to advance.


marcos-iov (approved):

nathanieliov (approved):

nathanieliov:
Rebase and then merge it

PR #30, opened 2024-01-11
Replaces the createPeginV1TxData with the one from pegin-address-veri…

jeremy-then:
Replaces the createPeginV1TxData with the one from pegin-address-verifier.

Since the `createPeginV1TxData` is already in the `pegin-address-verifier` repo, there's no need to keep it in the RITs.
This pr removes the one in the RITs and uses the one in `pegin-address-verifier`  instead.

Also, since I realized that most of the functions in the `2wp-utils-legacy.js` file where not being used, I removed all of them. There's only one last function in that file being used outside of it that we need to remove in order to delete this legacy file. A ticket has been created already to handle it.


marcos-iov (commented):
Nice! Well done

marcos-iov (approved):

nathanieliov (approved):

PR #32, opened 2024-02-12
Adds log4js. Updates wait for port attempts.

jeremy-then:
1) Adds log4js so we can add logs and specify when to run certain log level.
2) Updates `numRetries ` and `retryInterval`, `retryInterval` so it doesn't try to check if the port is available that often because it will probably not be, and reducing `numRetries` because since `retryInterval` has increased, no need for `numRetries` to be that big. This will help mitigate some random issues about not being able to find a port for bitcoind or a federate node in time.

### DIFF
--- a/tests/01_03_54-post-papyrus_coinbase_information.js
+++ b/tests/01_03_54-post-papyrus_coinbase_information.js
@@ -8,12 +8,14 @@ const { getBtcClient } = require('../lib//btc-client-provider');
 const btcEthUnitConverter = require('@rsksmart/btc-eth-unit-converter');
 const { getRskTransactionHelpers } = require('../lib/rsk-tx-helper-provider');
 
-let rskTxHelpers;
+const { getLogger } = require('../logger');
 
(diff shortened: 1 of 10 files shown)


nathanieliov (approved):

tests/01_03_54-post-papyrus_coinbase_information.js
```diff
@@ -8,12 +8,14 @@ const { getBtcClient } = require('../lib//btc-client-provider');
 const btcEthUnitConverter = require('@rsksmart/btc-eth-unit-converter');
 const { getRskTransactionHelpers } = require('../lib/rsk-tx-helper-provider');
 
-let rskTxHelpers;
+const { getLogger } = require('../logger');
 
 describe('Calling coinbase information methods after papyrus', () => {
 
+    const logger = getLogger();
```
marcos-iov, line 15: is this used anywhere?

marcos-iov (approved):

PR #45, opened 2024-06-10
Initial version from github actions workflow

fmacleal:
In order to be able to easily have more external contributors, we need to migrate some private pipelines running on jenkins to Github Actions.

This PR starts this migration, executing the steps necessary to run the integration tests between powpeg-node and rskj inside Github actions.


.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
```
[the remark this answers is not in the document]
fmacleal: For sure isn,t I will clean it up this PR. It's on draft and everything here will change. ;)

.github/images/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# -- java ---------------------------------------------------------
+ENV JAVA_VERSION 11
+
+RUN apt-get  -y install "openjdk-$JAVA_VERSION-jdk"
+
+ENV JAVA_HOME=/opt/java/openjdk \
+    PATH="/opt/java/openjdk/bin:$PATH"
+
+# -- python ---------------------------------------------------------
+ENV PYTHON_VERSION=2.7.16
```
jeremy-then, line 29: Are we going to use HSM1 or HSM2 here? For the first phase, we could simply ignore any HSM related stuff and use all key files. What do you think?

fmacleal: I understand this, but I thought that it wouldn't hurt have it here. Since we had the effort to add support on the image, and that maybe we could use this image to build and run full integration tests with Python, what do you think?

.github/images/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+set -e
+
+
+echo -e "\n\n--------- Starting the configuration of rskj ---------\n\n"
+cd /usr/src/
+git clone https://github.com/rsksmart/rskj.git rskj
+cd rskj && git checkout $RSKJ_BRANCH
+chmod +x ./configure.sh && chmod +x gradlew
+./configure.sh
+./gradlew --no-daemon clean build -x test
```
jeremy-then, line 12: I think there's no need to build the rskj project. The federate-node project will build it automatically while the federate-node is building.

.github/images/rit-local-configs/regtest-all-keys.js
```diff
@@ -0,0 +1,116 @@
+const path = require('path');
+const nodesConfigPath = 'config/node-configs';
+const keysPathResolve = 'node-keys';
+const classpath = process.env.POWPEG_NODE_JAR_PATH;
+const federatesLogbackPath = path.resolve(__dirname, 'logback');
```
jeremy-then, line 5:
```suggestion
const federatesLogbackPath = path.resolve(__dirname, 'logbacks');
```

jeremy-then: Or, change the `logbacks` part of the `.github/images/rit-local-configs/logbacks/logback-fed-5.xml` path to look like `.github/images/rit-local-configs/logback/logback-fed-5.xml`. It looks like it has an extra 's' in it.

.github/action.yml
```diff
@@ -0,0 +1,36 @@
...
+    description: 'The output message of the integration tests'
+
+runs:
+  using: docker
+  image: ./images/Dockerfile
+  env:
+    RSKJ_BRANCH: ${{ inputs.rskj-branch }}
+    FEDERATOR_BRANCH: ${{ inputs.federator-branch }}
+    RIT_BRANCH: ${{ inputs.rit-branch }}
+    RIT_LOG_LEVEL: ${{ inputs.rit-log-level }}
```
jeremy-then, line 36: Leave a new line at the end of the file.

.github/images/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+RUN chmod +x /usr/src/entrypoint.sh  \
+    && chmod +x /usr/src/configure_gradle_federator.sh \
+    && chmod +x /usr/src/configure_rit_locally.sh \
+    && mkdir -p /usr/src/bitcoindata \
+    && chmod -R 755 /usr/src/bitcoindata \
+    && ls -la /usr/src
+
+ENTRYPOINT ["/usr/src/entrypoint.sh"]
+
+EXPOSE 18332
```
jeremy-then, line 65: Leave a new line at the end of the file.

.github/images/rit-local-configs/regtest-all-keys.js
```diff
@@ -0,0 +1,116 @@
...
+                'federator.signers.BTC.path': path.resolve(__dirname, `${keysPathResolve}/reg5.key`),
+                'federator.signers.RSK.type': 'keyFile',
+                'federator.signers.RSK.path': path.resolve(__dirname, `${keysPathResolve}/reg5.key`),
+                'federator.signers.MST.type': 'keyFile',
+                'federator.signers.MST.path': path.resolve(__dirname, `${keysPathResolve}/reg5.key`)
+            },
+            nodeId: '6bb9eab797eadc8b697f0e82a01d01cabbfaaca37e5bafc06fdc6fdd38af894a9a8cbaf526d344b5df39b80433609e006586050fd2188d30ab000b0fb6a6baaf'
+        }
+    ]
+}
```
jeremy-then, line 116: Leave a new line at the end of the file.

.github/images/scripts/configure_gradle_federator.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Concatenate modifier and versionNumber
+FED_VERSION="$modifier-$versionNumber"
+echo "$FED_VERSION"
+
+if [[ $FED_VERSION == SNAPSHOT* || $FED_VERSION == RC* ]]; then
+  echo -e "$SETTINGS_GRADLE_CONTENT_LOCAL" > DONT-COMMIT-settings.gradle
+else
+  echo "Settings gradle script local not necessary"
+fi
```
jeremy-then, line 45: Leave a new line at the end of the file.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+        uses: docker://ghcr.io/rsksmart/rootstock-integration-tests/rit:latest
+        with:
+          rskj-branch: master
+          federator-branch: master
+
+      - name: Print RIT Status and Message
+        id: output
+        run: |
+          echo "RIT Status = ${{ steps.test-rit-action.outputs.status }}"
+          echo "RIT Message = ${{ steps.test-rit-action.outputs.message }}"
```
jeremy-then, line 106: Leave a new line at the end of the file.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+
+    steps:
+      - name: GitHub container registry login
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Test  RIT Action
```
jeremy-then, line 95:
```suggestion
      - name: Test RIT Action
```

.github/images/Dockerfile
jeremy-then: No need for them now. I see you removed it. 👍🏽

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
+name: RIT Action Continuous Integration Test
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
```
jurajpiar, line 7: and `*-rc` perhaps?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Test the RIT Container Action
+        id: test-container
+        env:
+          RSKJ_BRANCH: 'master'
+          FEDERATOR_BRANCH: 'master'
+        run: |
+          docker run \
+            --env RSKJ_BRANCH="${{ env.RSKJ_BRANCH }}" \
+            --env FEDERATOR_BRANCH="${{ env.FEDERATOR_BRANCH }}" \
```
jurajpiar, line 58: 1. what's the point of these vars?
```suggestion
        run: |
          docker run \
            --env RSKJ_BRANCH="master" \
            --env FEDERATOR_BRANCH="master" \
```

2. is it suppose to be only master? shouldn't we test it on the PRs? I think I'm missing something...

.gitignore
```diff
@@ -7,3 +7,5 @@ config/*.js
 logs
 .env
 .DS_Store
+.idea/
```
jurajpiar, line 10: For personal files, such as editor configs, use global `~/.gitignore` of configure rules for specific dirs in `~/.gitconfig`

jurajpiar (approved):
Cool stuff

.github/images/rit-local-configs/regtest-all-keys.js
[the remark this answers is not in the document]
jurajpiar: actually, I only just realised that the public keys are repeating, as well, so that could be reduced, too.

.github/images/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+echo -e "\n\n--------- RIT Tests Result ---------\n\n"
+if [ $STATUS -ne 0 ]; then
+  MESSAGE="Rootstock Integration Tests Status: FAILED"
+else
+  MESSAGE="Rootstock Integration Tests Status: PASSED"
+fi
+echo -e "$MESSAGE"
+
+echo "status=$STATUS" >> $env:GITHUB_OUTPUT
+echo "message=$MESSAGE" >> $env:GITHUB_OUTPUT
```
jurajpiar, line 51:
```suggestion
echo "status=$STATUS" >> $GITHUB_ENV
echo "message=$MESSAGE" >> $GITHUB_ENV
```

.github/images/entrypoint.sh
nathanieliov: Agree with @jeremy-then

.github/images/rit-local-configs/regtest-all-keys.js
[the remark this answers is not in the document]
nathanieliov: Very nice idea @jurajpiar

.github/images/entrypoint.sh
ATarrio: You should be able to set the output so that it can be picked up in a following step on the workflow, but you need to reference the GITHUB_OUTPUT variable as ${GITHUB_OUTPUT}
The curly braces shouldn't be mandatory, but it's always better to use them in bash.

.github/images/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+set -e
+
+
+echo -e "\n\n--------- Starting the configuration of rskj ---------\n\n"
+cd /usr/src/
+git clone https://github.com/rsksmart/rskj.git rskj
+cd rskj && git checkout $RSKJ_BRANCH
```
ATarrio, line 9:
```suggestion
cd rskj && git checkout ${INPUT_RSKJ_BRANCH}
```
If you look at the action run, this variable is being passed to the container as INPUT_RSKJ_BRANCH.
I suspect the command si still working now because it's just running 'git checkout'  without parameters.

.github/images/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+chmod +x ./configure.sh && chmod +x gradlew
+./configure.sh
+./gradlew --no-daemon clean build -x test
+
+echo -e  "\n\n--------- Starting the configuration of powpeg ---------\n\n"
+cd /usr/src/
+git clone https://github.com/rsksmart/powpeg-node.git powpeg
+cp configure_gradle_federator.sh powpeg
+cd powpeg && git checkout $FEDERATOR_BRANCH
```
ATarrio, line 18:
```suggestion
cd powpeg && git checkout ${INPUT_FEDERATOR_BRANCH}
```
Same as above, might want to use INPUT_FEDERATOR_BRANCH.

.github/workflows/ci.yml
fmacleal: To configure the branch using the actions input variable. By default, we run using master, but imagine that you want to test from a branch, or from a tag, you just call the action setting it and the test will run on that branch or tag.

.github/images/entrypoint.sh
fmacleal: Ok, will test this then. Let's see how it goes. 🙏

.github/images/entrypoint.sh
fmacleal: Good catch, will try this! Thank you so much! 🙏

.github/images/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+# -- nodeJs ---------------------------------------------------------
+ENV NODE_VERSION  v20.14.0
+RUN mkdir -p /usr/local/nvm
+ENV NVM_DIR /usr/local/nvm
+
+RUN  curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash \
+     && . $NVM_DIR/nvm.sh \
+     &&  nvm install $NODE_VERSION \
+    && nvm alias default $NODE_VERSION \
+    && nvm use default
```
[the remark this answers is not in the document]
fmacleal: Makes sense, I will try to download node directly.

.github/workflows/ci.yml
fmacleal: 1. This var it's making this container-action support configuration from the branch of RSKJ and POWPEG that you want to validate with your action. Imagine that you are executing the action built here by this CI on RSKJ, you can use a variable on this input executing the RIT on the branch or tag passed through this variable.

2. It can be anything, can be RC, tags, branches...I am creating the action flexible enough to support any input to be used on the checkout. Be aware that this CI.yml, it's responsible to build and push the `RIT container action` this action will be the one added on the powpeg and rskj pipelines.

jurajpiar: no, sorry I didn't explain myself very well. What I meant was that why hardcode an env var to the step just to use is as var next. Could you not just use the hardcoded values directly?

.github/images/rit-local-configs/regtest-all-keys.js
fmacleal: Applied your suggestion. Thanks.

ATarrio (approved):
LGTM

.github/images/rit-local-configs/logbacks/logback-fed-1.xml
```diff
@@ -0,0 +1,114 @@
...
+            <pattern>%date{yyyy-MM-dd-HH:mm:ss.SSSS} %p [%c{1}]  %m%n</pattern>
+        </encoder>
+        <!-- <filter class="ch.qos.rit-local-configs.classic.filter.ThresholdFilter">
+            <level>DEBUG</level>
+        </filter> -->
+    </appender>
+
+    <appender name="FILE-AUDIT"
+              class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>/Users/fredrhae/Documents/repos/powpeg-config/logs/rsk-fed-1.log</file>
```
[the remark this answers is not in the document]
fmacleal: I just replicated what needed to work locally, @jeremy-then has another PR to add them on the RIT logbacks folder directly. Once it's merged, we can clean it up this one. For now, let's let it as it is for the sake of simplicity, sounds good?

.github/images/rit-local-configs/regtest-all-keys.js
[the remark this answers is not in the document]
fmacleal: Since this refactor wasn't working, I haven't applied, we can try to implement it properly later.

lib/rsk-utils-legacy.js
```diff
@@ -26,7 +26,7 @@ var waitForSync = (rskClients) => {
   });
 };
 
-var waitForBlock = (rskClient, blockNumber, waitTime = 200, maxAttempts = 600) => {
+var waitForBlock = (rskClient, blockNumber, waitTime = 600, maxAttempts = 1000) => {
```
fmacleal, line 29: @jeremy-then, I had to increase it in some tests, take a look if I haven't increased too much. You said that for some other tests, these values couldn't be that high. 🙏

jurajpiar (approved):

lib/rsk-utils-legacy.js
jeremy-then: If the tests are running now for the environment these tests are running, that's fine.
Let's consider putting these values in the configuration, like in the `.env` file. What do you think? Could this be done in the scope of this pr? If not, let's create a ticket to handle that later.

jeremy-then (dismissed):
LGTM. Left a small optional comment.

lib/rsk-utils-legacy.js
fmacleal: Addressed, can you take a look @jeremy-then?

jurajpiar (approved):

lib/rsk-utils-legacy.js
jeremy-then: Looking good. The names look a bit general, what about specifying them a little more? Like:

```
waitForBlockAttemptTimeMillis = process.env.WAIT_FOR_BLOCK_ATTEMPT_TIME_MILLIS || 600
```

```
waitForBlockMaxAttempts = process.env. WAIT_FOR_BLOCK_MAX_ATTEMPTS || 1000;
```

And let's reduce the defualt values to something like `200` for the waiting and `160` for the attempts, since in a local computer having such a high value won't do any good, but slow down the tests.

Also, we can add a brief description comment above these new env variables in the `.env-example` file saying why we use them and need to update them depending on the environment.

```
# Amount of time in milliseconds for the `waitForBlock` utility function to wait. Configurable because depending on the resources of the machine the tests are running on, the wait time might change. In a machine with little resources (CPU, RAM, disk), a small wait time might not be enough because blocks can be mined slow and  `waitForBlock` might fail with a message like `Block number 800 never reached, last seen was 600`, or `Blockchain not advancing after attempting to find a new block 80 times checking every 200 milliseconds. Couldn't reach block number 800. Last block number seen was: 600`. In a machine with enough resources having a high wait time might be a waste of time since the tests would run slower because if this wait time. In this case, it can be set to a small value. `200` recommended for most machines with enough resources. `500`, `600`, etc., or more for machine with limited resources. Adjust as needed, starting with low values so the tests run as fast as they can.
WAIT_FOR_BLOCK_ATTEMPT_TIME_MILLIS=

# Max attempts for the `waitForBlock` utility function to 'wait' for the given block, trying to find that block once every `WAIT_FOR_BLOCK_ATTEMPT_TIME_MILLIS`. The same as the `WAIT_FOR_BLOCK_ATTEMPT_TIME_MILLIS` variable, the value for this variable could be updated depending on the machine the tests are running on. `80` recommended for most machines with enough resources. `160`, `250` or more for machine with limited resources. Adjust as needed, starting with low values so the tests run as fast as they can.
WAIT_FOR_BLOCK_MAX_ATTEMPTS=
```

If the descriptions are too long, we could move them to the `README.md` file. But I think it's important to be as detailed as possible, so we don't have other developers guessing why the test failed with that message.

Sorry that I was not clearer when I proposed the changes.

fmacleal: Followed and applied the suggestions.

jeremy-then (approved):
LGTM.

.github/images/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+RUN apt-get  -y install "openjdk-$JAVA_VERSION-jdk"
+
+ENV JAVA_HOME="/usr/lib/jvm/java-$JAVA_VERSION-openjdk-amd64" \
+    PATH="$JAVA_HOME:$PATH"
+
+RUN export JAVA_HOME=$JAVA_HOME  \
+    && export PATH="$JAVA_HOME:$PATH"
+
+# -- bitcoind ---------------------------------------------------------
+ENV BITCOIN_VERSION 0.18.1
```
lucasvuotto, line 36: We're running Bitcoin 24 in prod. Do we want 0.18.1 in here?

lib/rsk-utils-legacy.js
```diff
@@ -2,6 +2,8 @@ const expect = require('chai').expect;
 const btcClientProvider = require('./btc-client-provider');
 const { sequentialPromise, wait, interval } = require('./utils');
 const { getBridgeAbi, getLatestActiveForkName } = require('./precompiled-abi-forks-util');
+const waitForBlockAttemptTimeMillis = process.env.WAIT_FOR_BLOCK_ATTEMPT_TIME_MILLIS || 200
```
lucasvuotto, line 5: For consistency with the surrounding lines.

```suggestion
const waitForBlockAttemptTimeMillis = process.env.WAIT_FOR_BLOCK_ATTEMPT_TIME_MILLIS || 200;
```

.github/images/Dockerfile
jeremy-then: @lucasvuotto , unfortunately, yes. Bitcoind 24 doesn't allow us certain flexibilities easily, like programmatically creating a wallet to spend from, etc.
I remember working on an story to figure out if we could use Bitcoind 24 for the rits and we were unable to get it.
We should probably analyze that later again.
We might have the result of that analysis somewhere. I will get back to you if I get it.

fmacleal: Understood. Let's create a ticket for it and settle. :)

jeremy-then (commented):
LGTM.

jeremy-then (dismissed):
LGTM.

.github/action.yml
```diff
@@ -0,0 +1,36 @@
...
+
+inputs:
+  rskj-branch:
+    description: 'The rskj branch to checkout'
+    required: false
+    default: 'master'
+  federator-branch:
+    description: 'The federate-node branch to checkout'
+    required: false
+    default: 'master'
```
lucasvuotto, line 13: Purely cosmetic, but I prefer using `powpeg-node` instead of federator in here.

.github/images/Dockerfile
lucasvuotto: @jeremy-then can you share me that story? I'm willing to help with it.

.github/images/entrypoint.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+set -e
+
+echo -e "\n\n--------- Starting the configuration of rskj ---------\n\n"
+cd /usr/src/
+git clone https://github.com/rsksmart/rskj.git rskj
+cd rskj && git checkout ${INPUT_RSKJ_BRANCH}
```
lucasvuotto, line 8: This is an untrusted input. As such, please quote it.

```suggestion
cd rskj && git checkout "${INPUT_RSKJ_BRANCH}"
```

.github/images/entrypoint.sh
```diff
@@ -0,0 +1,50 @@
...
+chmod +x ./configure.sh && chmod +x gradlew
+./configure.sh
+
+echo -e  "\n\n--------- Starting the configuration of powpeg ---------\n\n"
+cd /usr/src/
+git clone https://github.com/rsksmart/powpeg-node.git powpeg
+cp configure_gradle_federator.sh powpeg
+cd powpeg && git checkout ${INPUT_FEDERATOR_BRANCH}
```
lucasvuotto, line 16: This is an untrusted input. As such, please quote it.

```suggestion
cd powpeg && git checkout "${INPUT_FEDERATOR_BRANCH}"
```

.github/images/entrypoint.sh
```diff
@@ -0,0 +1,50 @@
...
+./gradlew  --info --no-daemon clean build -x test
+
+echo -e "\n\n--------- Starting the configuration of RIT ---------\n\n"
+cd /usr/src/
+git clone https://github.com/rsksmart/rootstock-integration-tests.git rit
+mv configure_rit_locally.sh rit
+mv regtest.js rit/config/regtest.js
+mv /usr/src/logbacks/* /usr/src/rit/logbacks/
+cd rit
+git checkout ${INPUT_RIT_BRANCH}
```
lucasvuotto, line 30: This is an untrusted input. As such, please quote it.

```suggestion
git checkout "${INPUT_RIT_BRANCH}"
```

.github/images/entrypoint.sh
```diff
@@ -0,0 +1,50 @@
...
+chmod +x ./configure.sh
+./configure.sh
+./configure_rit_locally.sh "${FED_VERSION}"
+export LOG_LEVEL=${INPUT_RIT_LOG_LEVEL}
```
lucasvuotto, line 34: This is an untrusted input. As such, please quote it.

```suggestion
export LOG_LEVEL="${INPUT_RIT_LOG_LEVEL}"
```

Or use it in 2 steps, which gives the same guaruantees as quoting:

```
LOG_LEVEL=${INPUT_RIT_LOG_LEVEL}
export LOG_LEVEL
```

.github/images/entrypoint.sh
```diff
@@ -0,0 +1,50 @@
...
+echo -e "\n\n--------- RIT Tests Result ---------\n\n"
+if [ $STATUS -ne 0 ]; then
+  MESSAGE="Rootstock Integration Tests Status: FAILED"
+else
+  MESSAGE="Rootstock Integration Tests Status: PASSED"
+fi
+echo -e "$MESSAGE"
+
+echo "status=${STATUS}" >>${GITHUB_OUTPUT}
+echo "message=${MESSAGE}" >> ${GITHUB_OUTPUT}
```
lucasvuotto, line 50: In general, unless you're 100% sure about what you're doing and you fully control the value of a variable, please quote it when used.

```suggestion
echo "status=${STATUS}" >>"${GITHUB_OUTPUT}"
echo "message=${MESSAGE}" >>"${GITHUB_OUTPUT}"
```

lucasvuotto (changes requested):

.github/images/Dockerfile
diego-jeronymo, line 1: Hi @fmacleal, If the application doesn't need to run the container using the root user, I recommend explicitly creating a dedicated user to run the container.
e.g.
```
+ RUN useradd -m appuser
+ USER appuser 
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,101 @@
...
+      - name: GitHub container registry login
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Test RIT Action
+        id: test-rit-action
+        uses: docker://ghcr.io/rsksmart/rootstock-integration-tests/rit:latest
```
lucasvuotto, line 94:
```suggestion
        uses: docker://${{ env.LATEST_TAG }}:latest
```

container-action/scripts/configure_rit_locally.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+POWPEG_VERSION="${1:-SNAPSHOT-6.3.0.0}"
```
lucasvuotto, line 3: Do we really want this script to be run if no parameter is given? Hardcoding it here means either having to keep it up-to-date to avoid running in the edge case were no parameter is given, having SNAPSHOT-6.3.0.0 being used and scratching our heads about the results.

I'd prefer to be explicit and fail if no parameter is given. You can implement that like

```suggestion
if [ $# -ne 1 ] || [ -z "${1:-}" ]; then
    echo "Usage: ${0##*/} POWPEG_VERSION" >&2
    exit 1
fi
POWPEG_VERSION=$1
shift
```

container-action/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+     && . $NVM_DIR/nvm.sh \
+     &&  nvm install $NODE_VERSION \
+    && nvm alias default $NODE_VERSION \
+    && nvm use default
+
+ENV NODE_PATH $NVM_DIR/$NODE_VERSION/lib/node_modules
+ENV PATH $NVM_DIR/versions/node/$NODE_VERSION/bin:$PATH
+
+# -- java ---------------------------------------------------------
+ENV JAVA_VERSION 11
```
Vovchyk, line 25: could it be java17 pls?

fmacleal: Done and the test passed.

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

bcodesido (approved):
LGTM

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

container-action/scripts/configure_rit_locally.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+if [ $# -ne 1 ] || [ -z "${1:-}" ]; then
+    echo "Usage: ${0##*/} POWPEG_VERSION" >&2
+    exit 1
+fi
+POWPEG_VERSION=$1
+shift
+echo "POWPEG_VERSION received as parameter: $1"
```
lucasvuotto, line 9:
```suggestion
echo "POWPEG_VERSION received as parameter: $POWPEG_VERSION"
```

The `shift` changes the value of `$1`. In particular, given that the script aborts if it doesn't have exacly one parameter, this is accessing an unset variable. Depending on how it's run, that can be an error or a warning.

container-action/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+
+# -- java ---------------------------------------------------------
+ENV JAVA_VERSION 17
+
+RUN apt-get  -y install "openjdk-$JAVA_VERSION-jdk"
+
+ENV JAVA_HOME="/usr/lib/jvm/java-$JAVA_VERSION-openjdk-amd64"
+
+RUN export JAVA_HOME="$JAVA_HOME"  \
+    && export PATH="$JAVA_HOME:$PATH"
```
lucasvuotto, line 32: This is a no-op and adds an extra layer to the container. The reach of those `export` are for the whole `RUN` instruction, not for all the ones after. My understanding is that the previous `ENV JAVA_HOME=...` should be enough.

nathanieliov:
pipeline:run

bcodesido (approved):
LGTM

lucasvuotto (approved):

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

nathanieliov:
pipeline:run

container-action/README.md
```diff
@@ -0,0 +1,59 @@
...
+and should be executed before any release of both projects or any merge to the master/main branch. 
+
+To achieve this and make this test more accessible, we created a container-action created to execute this test, 
+it offers the flexibility to run the tests with any specific tag or branch from *powpeg-node* or *rskj*. 
+That way, we will add steps on each repository to run the integration tests with the version that we want to test. 
+No matter if it's a tag, a branch or a specific commit.
+
+## Inputs
+By default, all the inputs are pointed to the `master/main` branch of the repositories. But, ideally, we will adapt
+the action to receive the branches  received to be the one in a branch or tag that we want to test.
```
marcos-iov, line 18: This last part is a bit confusing

container-action/README.md
```diff
@@ -0,0 +1,59 @@
...
+The rskj branch to checkout. If no branch or tag passed, it will be used the default `master`.
+
+### `powpeg-node-branch`
+
+The powpeg-node branch to checkout. If no branch or tag passed, it will be used the default `master`.
+
+### `rit-branch`
+
+**Optional** The rootstock-integration-tests branch to checkout. This one it's optional, because it will be
+very unlikely that we need to use a different branch for the rootstock-integration-test. It's offered the possibility
```
marcos-iov, line 31: I wouldn't say it's very unlikely. Sometimes a new feature requires a change in the RIT so we need to use a specific RIT branch to run the tests along with the feature changes in rskj/powpeg.

Also, when there is a change to the RIT code and a PR against master we also want the tests to run with the RIT branch that has the changes included

container-action/rit-local-configs/logbacks/logback-fed-1.xml
```diff
@@ -0,0 +1,113 @@
...
+    <!-- Errors were reported during translation. -->
+    <!-- Could not find transformer for org.apache.log4j.rolling.RollingFileAppender -->
+    <appender name="stdout" class="ch.qos.logback.core.ConsoleAppender">
+        <Target>System.out</Target>
+        <encoder>
+            <pattern>%date{yyyy-MM-dd-HH:mm:ss.SSSS} %p [%c{1}]  %m%n</pattern>
+        </encoder>
+        <!-- <filter class="ch.qos.rit-local-configs.classic.filter.ThresholdFilter">
+            <level>DEBUG</level>
+        </filter> -->
```
marcos-iov, line 20: Should we keep this and other commented out parts?

lib/rsk-utils-legacy.js
```diff
@@ -2,8 +2,10 @@ const expect = require('chai').expect;
 const btcClientProvider = require('./btc-client-provider');
 const { sequentialPromise, wait, interval } = require('./utils');
 const { getBridgeAbi, getLatestActiveForkName } = require('./precompiled-abi-forks-util');
+const waitForBlockAttemptTimeMillis = process.env.WAIT_FOR_BLOCK_ATTEMPT_TIME_MILLIS || 200;
+const waitForBlockMaxAttempts = process.env. WAIT_FOR_BLOCK_MAX_ATTEMPTS || 160;
```
marcos-iov, line 6: The tests are failing when executed against this branch. I think it might be related to this, or maybe it's the default values set in `configure_rit_locally.sh` file.

Could you set them back to the original values please?

.gitignore
```diff
@@ -7,3 +7,4 @@ config/*.js
...
```
marcos-iov, line 10: We should a blank line at the end of this and other files that github is reporting

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

marcos-iov (approved):

PR #77, opened 2024-09-09
Increase locking cap to 21 million as part of the setup start

jeremy-then:
Increase locking cap to 21 million as part of the setup start

### DIFF
--- a/lib/rsk-utils.js
+++ b/lib/rsk-utils.js
@@ -333,26 +333,40 @@ const triggerRelease = async (rskTransactionHelpers, btcClient, callbacks = {}) 
  * Executes a method 'call' and calls the callback with the result of the call, then calls 'send' and waits for the transaction receipt to be available and returns it
  * @param {RskTransactionHelper} rskTxHelper to make transactions to the rsk network
  * @param {web3.eth.Contract.ContractSendMethod} method contract method to be invoked
- * @param {function} checkCallback callback to check the result of the method 'call' before calling 'send'
  * @param {string} from rsk address to send the transaction from
+ * @param {function} checkCallback callback to check the result of the method 'call' before calling 'send'
+ * @param {object} options optional. Object with the options for the transaction. It can contain the following properties: value, gasPrice, gas
  * @returns {web3.eth.TransactionReceipt} txReceipt
  */
-const sendTxWithCheck = async (rskTxHelper, method, from, checkCallback) => {
-
-  const callResult = await method.call({ from });
+const sendTxWithCheck = async (rskTxHelper, method, from, checkCallback, options = { value: 0, gasPrice: 0, gas: 0 }) => {
 
   if(checkCallback) {
+    const callResult = await method.call({ from });
     await checkCallback(callResult);
   }
 
-  const estimatedGas = await method.estimateGas({ from });
-  const txReceiptPromise = method.send({ from, value: 0, gasPrice: 0, gas: estimatedGas });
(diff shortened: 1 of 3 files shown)


tests/00_00_3-vote_for_locking_cap_to_21m.js
```diff
@@ -0,0 +1,59 @@
...
+            
+            nextIncrement = currentLockingCapValueInSatoshis * 2;
+
+            // Ensuring that the next increment is not greater than the target locking cap.
+            nextIncrement = Math.min(nextIncrement, targetLockingCapInSatoshis);
+
+            const increaseLockingCapCallResult = bridge.methods.increaseLockingCap(nextIncrement).send({ from: authAddress });
+            await wait(1000);
+            await rskUtils.mineAndSync(rskTxHelpers);
+            await increaseLockingCapCallResult;
```
marcos-iov, line 45: Not sure I follow this here, why create the call, then wait, then mine and finally execute the call. Shouldn't this be enough?

```suggestion
           await bridge.methods.increaseLockingCap(nextIncrement).send({ from: authAddress });
            await rskUtils.mineAndSync(rskTxHelpers);
```

jeremy-then: The test nodes don't mine automatically. If I do the `await bridge.methods.increaseLockingCap...`, it will get stuck for ever because it will try to get the transaction receipt but since it has not been mined, it will never return.

I will use the `sendTxWithCheck` function instead.

lib/rsk-utils.js
```diff
@@ -333,20 +333,22 @@ const triggerRelease = async (rskTransactionHelpers, btcClient, callbacks = {})
...
   if(checkCallback) {
+    const callResult = await method.call({ from });
     await checkCallback(callResult);
   }
 
-  const estimatedGas = await method.estimateGas({ from });
-  const txReceiptPromise = method.send({ from, value: 0, gasPrice: 0, gas: estimatedGas });
+  const { value, gasPrice, gas } = options;
+
+  const estimatedGas = gas ? gas : await method.estimateGas({ from });
```
marcos-iov, line 350: Check Sonar report here

lib/rsk-utils.js
```diff
@@ -333,20 +333,22 @@ const triggerRelease = async (rskTransactionHelpers, btcClient, callbacks = {})
...
- * @param {function} checkCallback callback to check the result of the method 'call' before calling 'send'
  * @param {string} from rsk address to send the transaction from
+ * @param {function} checkCallback callback to check the result of the method 'call' before calling 'send'
+ * @param {object} options optional. Object with the options for the transaction. It can contain the following properties: value, gasPrice, gas
  * @returns {web3.eth.TransactionReceipt} txReceipt
  */
-const sendTxWithCheck = async (rskTxHelper, method, from, checkCallback) => {
-
-  const callResult = await method.call({ from });
+const sendTxWithCheck = async (rskTxHelper, method, from, checkCallback, options = { value: 0, gasPrice: 0, gas: 0 }) => {
```
marcos-iov, line 341: Maybe receive options param first and checkCallback last? Mainly to avoid having to pass `null` when calling from `sendTransaction`

jeremy-then: Updated.

lib/rsk-utils.js
jeremy-then: Updated.

marcos-iov (approved):

PR #87, opened 2024-09-17
Adds new 2wp file starting with one pegin test checking event

jeremy-then:
This pr creates a new `lib/tests/2wp.js` file, with one initial pegin test, the most basic pegin.
This test checks the initial federation and sender address balances to compare them with the expected final balances.
This test also checks that the `pegin_btc` event is emitted with the expected values.
Asserts that the Bridge own balance and the Bridge utxos sum is as expected.


lib/constants.js
```diff
@@ -55,7 +55,14 @@ const PEGOUT_EVENTS = {
   BATCH_PEGOUT_CREATED: "batch_pegout_created",
   PEGOUT_TRANSACTION_CREATED: "pegout_transaction_created",
   PEGOUT_CONFIRMED: "pegout_confirmed"
-}
+};
+
+const PEGIN_EVENTS = {
```
marcos-iov, line 60: Not for now, but I would consider splitting this file before it grows too much. 2wpConstants, BrideEvents, FedChangeConstants, etc

marcos-iov (approved):

PR #89, opened 2024-09-18
Using 'parameterized tests' for 2wp.js pegin tests

jeremy-then:


lib/tests/2wp.js
```diff
@@ -29,6 +32,87 @@ const setupBridgeDonation = async (rskTxHelpers, btcTxHelper) => {
...
+  {
+    description: 'should do legacy pegin with one input and one output to the federation',
+    initialBtcSenderBalancesInBtc: [ MINIMUM_PEGIN_VALUE_IN_BTC ],
+    btcSenderAmountsToSendToFed: [ MINIMUM_PEGIN_VALUE_IN_BTC ],
+    outputsToFed: [ MINIMUM_PEGIN_VALUE_IN_BTC ],
+  },
+  {
+    description: 'should do legacy pegin with multiple inputs from different accounts and one output to the federation',
+    initialBtcSenderBalancesInBtc: [ MINIMUM_PEGIN_VALUE_IN_BTC, MINIMUM_PEGIN_VALUE_IN_BTC, MINIMUM_PEGIN_VALUE_IN_BTC ], // Each sender is funded with some amount
+    btcSenderAmountsToSendToFed: [ MINIMUM_PEGIN_VALUE_IN_BTC, MINIMUM_PEGIN_VALUE_IN_BTC, MINIMUM_PEGIN_VALUE_IN_BTC ], // Each sender decides how much to send to the federation
```
nathanieliov, line 105: `btcSenderAmountsToSendToFed` seems a bit complex name to me. Maybe `peginSenderAmounts` by including pegin, we can assume this refers to funds sent to fed. Wdyt?

lib/tests/2wp.js
```diff
@@ -29,6 +32,87 @@ const setupBridgeDonation = async (rskTxHelpers, btcTxHelper) => {
...
+
+const getTotalAmountToFed = (outputsToFed) => {
+  return outputsToFed.reduce((total, amount) => total + amount, 0);
+};
+
+const getFinalRskRecipientBalances = async (sendersInfo) => {
+  return await Promise.all(sendersInfo.map(senderInfo => rskTxHelper.getBalance(senderInfo.rskRecipientRskAddressInfo.address)));
+};
+
+const legacyPeginTestsHappyPath = [
```
nathanieliov, line 95: These are not simple legacy pegins. I consider these legacy pegins as border cases, as well as uncommon/complex legacy pegins, at least those ones sending multiple outputs to the fed, since there's no reason for wallet to do that.

That being said, I would add a prefix to this variable like `uncommon/complex-LegacyPeginTests` . wdyt?

PR #92, opened 2024-09-24
Adds 'should do a basic legacy pegin with the value above minimum' test

jeremy-then:
Adds a basic legacy pegin test with a value above minimum.
Besides the value above minimum, this pegin does the same checks as this test in this pr: https://github.com/rsksmart/rootstock-integration-tests/pull/87

### DIFF
--- a/lib/tests/2wp.js
+++ b/lib/tests/2wp.js
@@ -102,6 +102,48 @@ const execute = (description, getRskHost) => {
 
     });
 
+    it('should do a basic legacy pegin with the value above minimum', async () => {
+
+      // Arrange
+
+      const initialFederationAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, federationAddress);
+      const senderRecipientInfo = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
+      const initialSenderAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, senderRecipientInfo.btcSenderAddressInfo.address);
+      // Value above minimum
+      const peginValueInSatoshis = minimumPeginValueInSatoshis + Number(btcToSatoshis(0.1));
+
+      // Act
+
+      const btcPeginTxHash = await sendPegin(rskTxHelper, btcTxHelper, senderRecipientInfo.btcSenderAddressInfo, satoshisToBtc(peginValueInSatoshis));
+      await ensurePeginIsRegistered(rskTxHelper, btcPeginTxHash);
+
+      // Assert
+
+      // The btc pegin tx is already marked as processed by the bridge
+      const isBtcTxHashAlreadyProcessed = await bridge.methods.isBtcTxHashAlreadyProcessed(btcPeginTxHash).call();
+      expect(isBtcTxHashAlreadyProcessed).to.be.true;
+
+      // The pegin_btc event is emitted with the expected values
+      const recipient1RskAddressChecksumed = rskTxHelper.getClient().utils.toChecksumAddress(ensure0x(senderRecipientInfo.rskRecipientRskAddressInfo.address));
+      const expectedEvent = createExpectedPeginBtcEvent(PEGIN_EVENTS.PEGIN_BTC, recipient1RskAddressChecksumed, btcPeginTxHash, peginValueInSatoshis);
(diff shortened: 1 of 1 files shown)


lib/tests/2wp.js
```diff
@@ -95,6 +95,58 @@ const execute = (description, getRskHost) => {
...
+      // Act
+
+      const btcPeginTxHash = await sendPegin(rskTxHelper, btcTxHelper, senderRecipientInfo.btcSenderAddressInfo, satoshisToBtc(peginValueInSatoshis));
+      await ensurePeginIsRegistered(rskTxHelper, btcPeginTxHash);
+
+      // Assert
+
+      // The btc pegin tx is already marked as processed by the bridge
+      const isBtcTxHashAlreadyProcessed = await bridge.methods.isBtcTxHashAlreadyProcessed(btcPeginTxHash).call();
+      expect(isBtcTxHashAlreadyProcessed).to.be.true;
```
marcos-iov, line 119: Should this maybe be part of `ensurePeginIsRegistered`?

lib/tests/2wp.js
```diff
@@ -95,6 +95,58 @@ const execute = (description, getRskHost) => {
...
+
+      // The pegin_btc event is emitted with the expected values
+      const recipient1RskAddressChecksumed = rskTxHelper.getClient().utils.toChecksumAddress(ensure0x(senderRecipientInfo.rskRecipientRskAddressInfo.address));
+      const expectedEvent = createExpectedPeginBtcEvent(PEGIN_EVENTS.PEGIN_BTC, recipient1RskAddressChecksumed, btcPeginTxHash, peginValueInSatoshis);
+      const btcTxHashProcessedHeight = Number(await bridge.methods.getBtcTxHashProcessedHeight(btcPeginTxHash).call());
+      const peginBtcEvent = await findEventInBlock(rskTxHelper, expectedEvent.name, btcTxHashProcessedHeight);
+      expect(peginBtcEvent).to.be.deep.equal(expectedEvent);
```
marcos-iov, line 126: This could also go into it's own method, right?

lib/tests/2wp.js
jeremy-then: It depends. Somethings we want to assert that this is false. Inside `ensurePeginIsRegistered` it should always be true. But we can put it inside the `ensurePeginIsRegistered` and when needed, assert from the test that it is false.

lib/tests/2wp.js
```diff
@@ -95,6 +95,58 @@ const execute = (description, getRskHost) => {
...
 
+    it('should do a basic legacy pegin with the value above minimum', async () => {
+
+      // Arrange
+
+      const initialBridgeBalance = Number(await rskTxHelper.getBalance(BRIDGE_ADDRESS));
+      const initialBridgeUtxosBalance = await getBridgeUtxosBalance(rskTxHelper);
+      const initialFederationAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, federationAddress);
+      const senderRecipientInfo = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
+      const initialSenderAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, senderRecipientInfo.btcSenderAddressInfo.address);
```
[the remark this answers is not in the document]

lib/tests/2wp.js

lib/tests/2wp.js
```diff
@@ -115,6 +115,39 @@ const execute = (description, getRskHost) => {
...
 
+    it('should do a basic legacy pegin with the value above minimum', async () => {
+
+      // Arrange
+
+      const initial2wpBalances = await get2wpInitialBalances();
+      const senderRecipientInfo = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
+      const initialSenderAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, senderRecipientInfo.btcSenderAddressInfo.address);
+      // Value above minimum
+      const peginValueInSatoshis = minimumPeginValueInSatoshis + Number(btcToSatoshis(0.1));
```
marcos-iov, line 126:
```suggestion
      const peginValueInSatoshis = minimumPeginValueInSatoshis + 1;
```
Maybe make just above the minimum? To be sure

marcos-iov (approved):

PR #93, opened 2024-09-24
Adds legacy pegin test with multiple inputs from multiple accounts.

jeremy-then:
Adds 'should do legacy pegin with multiple inputs from different accounts and one output to the federation with value exactly minimum' test.

This tests what would happen if different bitcoin accounts create a pegin transaction where each of the inputs belong to a different account.

Asserts that the pegin is successful.

Asserts that the `pegin_btc` event is emitted with the expected values.

Asserts that only the first btc sender's input's derived rsk address is the one that receives the funds.

### DIFF
--- a/lib/btc-utils.js
+++ b/lib/btc-utils.js
@@ -5,7 +5,6 @@ const { retryWithCheck } = require('./utils');
 const { getLogger } = require('../logger');
 const { btcToSatoshis } = require('@rsksmart/btc-eth-unit-converter');
 const BtcTransactionHelper = require('btc-transaction-helper/btc-transaction-helper');
-const SpendableUtxosInformation = require('btc-transaction-helper');
 
 const logger = getLogger();
 
@@ -188,19 +187,6 @@ const waitForBitcoinMempoolToGetTxs = async (btcTxHelper, maxAttempts = 3, check
     return Number(btcToSatoshis(await btcTxHelper.getAddressBalance(btcAddress)));
   };
 
-  /**
-   * Gets the list of utxos that sum up to the `amountInBtc` for the given btc address, and includes the calculated change.
-   * Once the sum of the selected utxos is equal o greater than the `amountInBtc`, no more utxos will be added.
-   * This doesn't return the list of utxos from the account, only utxos that sum up to `amountInBtc`.
-   * @param {BtcTransactionHelper} btcTxHelper to make calls in the bitcoin network
-   * @param {string} address the btc account address to get the utxos from
-   * @param {number} amountInBtc the amount in btc the utxos need to sum up to
-   * @returns {Promise<import('btc-transaction-helper').SpendableUtxosInformation>} an object containing the list of `utxos` and the `change` difference of the sum of utxos minus the `amountInBtc`
-   */
-  const getAddressUtxosInfo = async (btcTxHelper, address, amountInBtc) => {
-    return await btcTxHelper.selectSpendableUTXOsFromAddress(address, amountInBtc);
-  };
-
   module.exports = {
     publicKeyToCompressed,
     fundAddressAndGetData,
@@ -208,5 +194,4 @@ const waitForBitcoinMempoolToGetTxs = async (btcTxHelper, maxAttempts = 3, check
     waitForBitcoinTxToBeInMempool,
     waitForBitcoinMempoolToGetTxs,
     getBtcAddressBalanceInSatoshis,
-    getAddressUtxosInfo,
   };
--- a/lib/btc-utils.js
+++ b/lib/btc-utils.js
@@ -4,6 +4,8 @@ const pmtBuilder = require('@rsksmart/pmt-builder');
 const { retryWithCheck } = require('./utils');
 const { getLogger } = require('../logger');
 const { btcToSatoshis } = require('@rsksmart/btc-eth-unit-converter');
+const BtcTransactionHelper = require('btc-transaction-helper/btc-transaction-helper');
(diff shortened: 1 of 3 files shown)


lib/tests/2wp.js
```diff
@@ -139,6 +207,34 @@ const assert2wpBalancesAfterSuccessfulPegin = async (initial2wpBalances, peginVa
 
 };
 
+const addInputs = (tx, senderUtxosInfo) => {
+  senderUtxosInfo.utxos.forEach(utxo => {
+    tx.addInput(Buffer.from(utxo.txid, 'hex').reverse(), utxo.vout);
+  });
+};
+
+const addChangeOutputs = (tx, changeAddress, changeInSatoshis) => {
```
nathanieliov, line 216: I think it should say `addChangeOutput` because it is just adding a single output. Wdyt?

lib/tests/2wp.js
```diff
@@ -139,6 +207,34 @@ const assert2wpBalancesAfterSuccessfulPegin = async (initial2wpBalances, peginVa
...
+  if(changeInSatoshis > 0) {
+    tx.addOutput(
+      bitcoinJsLib.address.toOutputScript(changeAddress, btcTxHelper.btcConfig.network),
+      changeInSatoshis - btcFeeInSatoshis
+    );
+  }
+};
+
+const ensurePeginIsPushed = async (btcPeginTxHash, expectedUtxosCount = 1) => {
```
nathanieliov, line 225: I would like this function to return the peginRskTxHash, wdyt?

nathanieliov (approved):
Just two minor comments

lib/tests/2wp.js
jeremy-then: Out of the scope. It would complicate things quite a bit.

lib/tests/2wp.js
jeremy-then: Thanks. I updated it in the next pr. https://github.com/rsksmart/rootstock-integration-tests/pull/94

nathanieliov: Nice

lib/btc-utils.js
```diff
@@ -4,6 +4,8 @@ const pmtBuilder = require('@rsksmart/pmt-builder');
 const { retryWithCheck } = require('./utils');
 const { getLogger } = require('../logger');
 const { btcToSatoshis } = require('@rsksmart/btc-eth-unit-converter');
+const BtcTransactionHelper = require('btc-transaction-helper/btc-transaction-helper');
+const SpendableUtxosInformation = require('btc-transaction-helper');
```
marcos-iov, line 8: where is this used?

lib/btc-utils.js
```diff
@@ -184,7 +186,20 @@ const waitForBitcoinMempoolToGetTxs = async (btcTxHelper, maxAttempts = 3, check
...
+  /**
+   * Gets the list of utxos that sum up to the `amountInBtc` for the given btc address, and includes the calculated change.
+   * Once the sum of the selected utxos is equal o greater than the `amountInBtc`, no more utxos will be added.
+   * This doesn't return the list of utxos from the account, only utxos that sum up to `amountInBtc`.
+   * @param {BtcTransactionHelper} btcTxHelper to make calls in the bitcoin network
+   * @param {string} address the btc account address to get the utxos from
+   * @param {number} amountInBtc the amount in btc the utxos need to sum up to
+   * @returns {Promise<import('btc-transaction-helper').SpendableUtxosInformation>} an object containing the list of `utxos` and the `change` difference of the sum of utxos minus the `amountInBtc`
+   */
+  const getAddressUtxosInfo = async (btcTxHelper, address, amountInBtc) => {
```
marcos-iov, line 200: I think this name is not very explanatory, by the name I thought it would return the list of available utxos for a given address.

Alos not sure this function makes much sense, it basically takes the 3 parameters passed and makes the call. What's the difference with calling btcTxHelper directly?

lib/tests/2wp.js
```diff
@@ -107,6 +107,74 @@ const execute = (description, getRskHost) => {
...
+      // Adding inputs
+      addInputs(tx, sender1UtxosInfo);
+      addInputs(tx, sender2UtxosInfo);
+
+      // Adding output to federation
+      addOutputToFed(tx, peginValueInSatoshis);
+
+      // Adding change outputs
+      addChangeOutputs(tx, senderRecipientInfo1.btcSenderAddressInfo.address, sender1ChangeInSatoshis);
+      addChangeOutputs(tx, senderRecipientInfo2.btcSenderAddressInfo.address, sender2ChangeInSatoshis);
```
marcos-iov, line 141: Can't this be a single method `addOutput(tx, amount, destination)`

lib/tests/2wp.js
```diff
@@ -139,6 +207,34 @@ const assert2wpBalancesAfterSuccessfulPegin = async (initial2wpBalances, peginVa
 
 };
 
+const addInputs = (tx, senderUtxosInfo) => {
```
marcos-iov, line 210:
```suggestion
const addInputs = (tx, utxos) => {
```
Maybe that simplifies? No need to receive the whole object

lib/btc-utils.js
jeremy-then: Updated.

lib/btc-utils.js
jeremy-then: Updated.

lib/tests/2wp.js

lib/tests/2wp.js

marcos-iov (approved):

PR #94, opened 2024-09-24
Adds legacy pegin test with multiple inputs and outputs.

jeremy-then:
Adds 'should do legacy pegin with multiple inputs from different accounts and two outputs to the federation with value exactly minimum' test.

This test checks the same things as https://rsklabs.atlassian.net/browse/POWPEG-1545, but sending multiple outputs to the federation as well, ensuring that everything else works as expected under this scenario.

Fixes bugs in `waitForBitcoinTxToBeInMempool`.
Adds balance check in `createSenderRecipientInfo` since sometimes the fund doesn't reach the mempool or doesn't get mined in time.
Renames some variables.

### DIFF
--- a/lib/2wp-utils.js
+++ b/lib/2wp-utils.js
@@ -107,8 +107,11 @@ const createPegoutRequest = async (rskTxHelper, amountInRBTC, requestSize = 1) =
  */
 const isUtxoRegisteredInBridge = async (rskTxHelper, peginTxHash, expectedUxtosCount = 1) => {
     const bridgeState = await getBridgeState(rskTxHelper.getClient());
-    return bridgeState.activeFederationUtxos
-        .filter(utxo =>  utxo.btcTxHash === peginTxHash).length === expectedUxtosCount;
+    const foundUtxos = bridgeState.activeFederationUtxos.filter(utxo =>  utxo.btcTxHash === peginTxHash);
+    if(foundUtxos.length > 0 && foundUtxos.length !== expectedUxtosCount) {
+        throw new Error(`Found ${foundUtxos.length} utxo(s) with the pegin tx hash ${peginTxHash} in the bridge, but expected to find ${expectedUxtosCount} utxo(s).`);
+    }
+    return foundUtxos.length === expectedUxtosCount;
 };
 
 const mineForPeginRegistration = async (rskTxHelper, btcTxHelper) => {
@@ -245,7 +248,17 @@ const createSenderRecipientInfo = async (rskTxHelper, btcTxHelper, type = 'legac
     await rskTxHelper.importAccount(rskRecipientRskAddressInfo.privateKey);
     await rskTxHelper.unlockAccount(rskRecipientRskAddressInfo.address);
     if(Number(initialAmountToFundInBtc) > 0) {
-        await btcTxHelper.fundAddress(btcSenderAddressInfo.address, initialAmountToFundInBtc);
+        const initialAddressBalance = await btcTxHelper.getAddressBalance(btcSenderAddressInfo.address);
+        const fundTxHash = await btcTxHelper.fundAddress(btcSenderAddressInfo.address, initialAmountToFundInBtc);
+        const finalAddressBalance = await btcTxHelper.getAddressBalance(btcSenderAddressInfo.address);
+        // If the final address balance is the same as the initial balance, then the tx has not yet reached the mempool or has not been mined. Rare condition that may happen randomly.
+        if(finalAddressBalance === initialAddressBalance) {
+            const inMempool = await waitForBitcoinTxToBeInMempool(btcTxHelper, fundTxHash);
+            if(inMempool) {
+                await btcTxHelper.mine();
+            }
+        }
+        
     }
     return {
         btcSenderAddressInfo,
--- a/lib/btc-utils.js
+++ b/lib/btc-utils.js
@@ -107,7 +107,7 @@ const waitForBitcoinTxToBeInMempool = async (btcTxHelper, btcTxHash, maxAttempts
       const isTxInMempool = bitcoinMempool.includes(btcTxHash);
       if(!isTxInMempool) {
         logger.debug(`[${waitForBitcoinTxToBeInMempool.name}::${bitcoinMempoolHasTx.name}] Attempting to check if the btc tx (${btcTxHash}) was already mined since it's not in the mempool yet.`);
-        const tx = await btcTransactionHelper.getTransaction(btcTxHash);
+        const tx = await btcTxHelper.getTransaction(btcTxHash);
         if(tx) {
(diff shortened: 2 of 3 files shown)


lib/2wp-utils.js
```diff
@@ -107,8 +107,11 @@ const createPegoutRequest = async (rskTxHelper, amountInRBTC, requestSize = 1) =
  */
 const isUtxoRegisteredInBridge = async (rskTxHelper, peginTxHash, expectedUxtosCount = 1) => {
     const bridgeState = await getBridgeState(rskTxHelper.getClient());
-    return bridgeState.activeFederationUtxos
-        .filter(utxo =>  utxo.btcTxHash === peginTxHash).length === expectedUxtosCount;
+    const foundUtxos = bridgeState.activeFederationUtxos.filter(utxo =>  utxo.btcTxHash === peginTxHash);
+    if(foundUtxos.length > 0 && foundUtxos.length !== expectedUxtosCount) {
+        throw new Error(`Found ${foundUtxos.length} utxo(s) with the pegin tx hash ${peginTxHash} in the bridge, but expected to find ${expectedUxtosCount} utxo(s).`);
```
jeremy-then, line 112: Throwing exception here so it doesn't fail silently. Otherwise it might be confusing for some tests.

lib/btc-utils.js
```diff
@@ -108,7 +107,7 @@ const waitForBitcoinTxToBeInMempool = async (btcTxHelper, btcTxHash, maxAttempts
       const isTxInMempool = bitcoinMempool.includes(btcTxHash);
       if(!isTxInMempool) {
         logger.debug(`[${waitForBitcoinTxToBeInMempool.name}::${bitcoinMempoolHasTx.name}] Attempting to check if the btc tx (${btcTxHash}) was already mined since it's not in the mempool yet.`);
-        const tx = await btcTransactionHelper.getTransaction(btcTxHash);
+        const tx = await btcTxHelper.getTransaction(btcTxHash);
```
jeremy-then, line 110: Bug 1 in this `waitForBitcoinTxToBeInMempool` function. It didn't have an `await`.

lib/btc-utils.js
```diff
@@ -136,7 +135,7 @@ const waitForBitcoinTxToBeInMempool = async (btcTxHelper, btcTxHash, maxAttempts
       throw e;
     };
   
-    const { result: btcTxAlreadyFoundInMempool } = retryWithCheck(
+    const { result: btcTxAlreadyFoundInMempool } = await retryWithCheck(
```
jeremy-then, line 138: Bug 2 in this `waitForBitcoinTxToBeInMempool` function. It didn't have an `await`.

lib/tests/2wp.js
```diff
@@ -175,13 +175,82 @@ const execute = (description, getRskHost) => {
...
+
+      const signedTx = await btcTxHelper.nodeClient.signTransaction(tx.toHex(), [], sendersPrivateKeys);
+
+      // Act
+
+      const btcPeginTxHash = await btcTxHelper.nodeClient.sendTransaction(signedTx);
+      // Assert
+
+      // Since we are not using `sendPegin` here, we need to do some extra steps before ensuring the pegin is registered.
+      const expectedCountOfThisPeginUtxosInTheBridge = 2;
```
jeremy-then, line 222: If we are expected 2 outputs to the federation in this pegin, we need to explicitly let the `ensurePeginIsPushed` know about it. Otherwise, it will fail if it finds a different number of utxos with this pegin btc hash.

nathanieliov (approved):

lib/2wp-utils.js
```diff
@@ -107,8 +107,11 @@ const createPegoutRequest = async (rskTxHelper, amountInRBTC, requestSize = 1) =
...
+    }
+    return foundUtxos.length === expectedUxtosCount;
```
marcos-iov, line 114: guess this will always be true then?

lib/2wp-utils.js
marcos-iov: Not sure about this, the method is called `isUtxoRegisteredInBridge`. I would expect it to return true or false. Now it's true or exception.

I understand the possible confusion. Perhaps it could be solved with a rename `assertUTXOsRegisteredInBridge`? Do nothing or throw exception

lib/2wp-utils.js
```diff
@@ -245,7 +248,17 @@ const createSenderRecipientInfo = async (rskTxHelper, btcTxHelper, type = 'legac
...
     if(Number(initialAmountToFundInBtc) > 0) {
-        await btcTxHelper.fundAddress(btcSenderAddressInfo.address, initialAmountToFundInBtc);
+        const initialAddressBalance = await btcTxHelper.getAddressBalance(btcSenderAddressInfo.address);
+        const fundTxHash = await btcTxHelper.fundAddress(btcSenderAddressInfo.address, initialAmountToFundInBtc);
+        const finalAddressBalance = await btcTxHelper.getAddressBalance(btcSenderAddressInfo.address);
+        // If the final address balance is the same as the initial balance, then the tx has not yet reached the mempool or has not been mined. Rare condition that may happen randomly.
+        if(finalAddressBalance === initialAddressBalance) {
+            const inMempool = await waitForBitcoinTxToBeInMempool(btcTxHelper, fundTxHash);
+            if(inMempool) {
+                await btcTxHelper.mine();
```
marcos-iov, line 258: What happens if we do this always? Instead of only doing it when initialBalance == finalBalance?

No big deal but I feel it would simplify the method

lib/2wp-utils.js
jeremy-then: Not really. In fact, this function gets called many times and returns `false` when the utxo is not yet in the Bridge. Which is ok. It's the `ensurePeginIsRegistered` function that calls the `isUtxoRegisteredInBridge` function many times.

When we register a pegin btc transaction in the Bridge, it could register either 1 or multiple utxos sharing the same btc transaction hash. If we are expecting 2 utxos from a pegin tx with the same btc transaction hash to be registered in the Bridge, but only 1 is registered, is it really successfully registered? Obviously not.
In a scenario like that, we want to throw an Exception.

lib/2wp-utils.js
jeremy-then: It will not. I replied to the comment above.

lib/2wp-utils.js
jeremy-then: It cannot be called `assertUTXOsRegisteredInBridge` because it would fail if it doesn't find the utxo in the Bridge. Which is not what we want. Because this function is called multiple times, until the uxto gets registered. Only after finding the utxo and the expected utxo count it will return true.

jeremy-then:
This pr includes a test that uses multiple inputs from different btc addresses. Sometimes this is a problem because while funding the second address, the wallet might choose the utxo of the first address to fund the second address, which leaves the first address without balance. This happens because the addresses are part of the same wallet and the wallet takes the utxos that it think is best. So, sometimes this test fails because of the first address not having balance.
Closing this pr until we find a better solution.

PR #96, opened 2024-09-24
Adds 'should reject a legacy pegin with the value exactly below minimum'

jeremy-then:
Adds a 'should reject a legacy pegin with the value exactly below minimum' test.

This test sends a pegin legacy with a value exactly below the minimum (the minimum pegin value minus 1 satoshis).

This tests is expected to not be marked as processed by the bridge and to emit a `rejected_pegin` event with the reason '5', which is 'invalid amount'.

With this test, we assert that the sender's derived rsk address does not have the funds, that the federation increases its funds but the funds are not refunded to the sender either.


jeremy-then:
New pr here: https://github.com/rsksmart/rootstock-integration-tests/pull/107

PR #98, opened 2024-09-25
Adds 'should reject a basic pegin v1 with value exactly below minimum…

jeremy-then:
Adds 'should reject a basic pegin v1 with value exactly below minimum' test.

This test sends a pegin v1 with a value exactly below minimum (the minimum pegin value minus 1 satoshis), to assert that it should be rejected with rejection reason 5 (invalid amount).

Asserts that the sender funds are lost and that the sender's derived rsk address nor the pegin v1 rsk recipient address get any funds, but the federation balance is increased by the pegin amount while the Bridge doesn't mark that pegin tx as processed.


jeremy-then:
https://github.com/rsksmart/rootstock-integration-tests/pull/108

PR #103, opened 2024-09-30
Adding repo owner input

rmoreliovlabs:


jeremy-then (dismissed):
LGTM.

fmacleal (approved):
looks good, straight to the point but working fine. Well done!

container-action/entrypoint.sh
fmacleal, line 1: suggestion:

Use the same `REPO_OWNER` in the line 26 for powpeg as well. Just to enable external contributors for them as well.

fmacleal (changes requested):
Just a small fix to use the `REPO_OWNER` for powpeg as well.

container-action/entrypoint.sh
rmoreliovlabs: Thanks for your input! I see where you're coming from, but I think using the same `REPO_OWNER` variable for both the forked `RSKJ` repo and the original Powpeg repo could lead to some confusion. If we set `REPO_OWNER` to the fork, we might inadvertently reference the wrong repository when working with Powpeg

It might make more sense to create a separate variable for Powpeg to avoid any conflicts. We could implement it similarly to how we handled it for RSKJ [here](https://github.com/rsksmart/rskj/pull/2774). What do you think?

fmacleal: Got it, if that's the case, I would suggest create a `REPO_OWNER` for each one. `REPO_OWNER_RSKJ` and `REPO_OWNER_POWPEG`. Then we could differentiate. And, I would suggest do in this PR so we can fix this issue only once.  Otherwise, to have external contributors, we will have to do a PR again for powpeg. 

But, it's up to the team decide, it can be discussed async on slack. This wouldn't prevent the approval, because this code works.

fmacleal (dismissed):
As a general rule of thumb, I will approve, but check my comment about do the fix for powpeg as well. I believe we could fix both issues so we allow external contributors for both.

container-action/entrypoint.sh
```diff
@@ -5,16 +5,18 @@ RSKJ_BRANCH="${INPUT_RSKJ_BRANCH}"
...
 RIT_BRANCH="${INPUT_RIT_BRANCH}"
 LOG_LEVEL="${INPUT_RIT_LOG_LEVEL}"
+REPO_OWNER="${INPUT_REPO_OWNER:-rsksmart}" # Default to 'rsksmart' if not provided
 
 echo -e "\n\n--------- Input parameters received ---------\n\n"
 echo "RSKJ_BRANCH=$RSKJ_BRANCH"
 echo "POWPEG_NODE_BRANCH=$POWPEG_NODE_BRANCH"
 echo "RIT_BRANCH=$RIT_BRANCH"
 echo "LOG_LEVEL=$LOG_LEVEL"
+echo "REPO_OWNER=$REPO_OWNER"
```
Vovchyk, line 15: @rmoreliovlabs , as external PRs can also be created against the PowPeg repo, we may also need to clone `powpeg-node` repo with `$REPO_OWNER`. And then fallback to `rsksmart`, if that fails, wdyt?

eg.:
```
git clone "https://github.com/$REPO_OWNER/rskj.git" rskj || git clone "https://github.com/rsksmart/rskj.git" rskj

git clone "https://github.com/$REPO_OWNER/powpeg-node.git" rskj || git clone "https://github.com/rsksmart/powpeg-node.git" rskj
```

Vovchyk (changes requested):

container-action/entrypoint.sh
Vovchyk: or even better, you could check if both: external repo + branch exists, eg.:

```
          IS_RSKJ_BRANCH=$(git ls-remote https://github.com/$REPO_OWNER/rskj.git "$RSKJ_BRANCH")
          if test -n "${IS_RSKJ_BRANCH}"; then
            echo "Found matching branch name in $REPO_OWNER/rskj.git repo"
            git clone "https://github.com/$REPO_OWNER/rskj.git" rskj
          else
            echo "Found matching branch name in rsksmart/rskj.git repo"
            git clone "https://github.com/rsksmart/rskj.git" rskj
          fi
```

rmoreliovlabs: Addressed, implemented your suggestion for both RSKJ and POWPEG repos, thanks!

Vovchyk: @rmoreliovlabs, that might not work if expected branch is missing in the counterpart forked repo. Check [this](https://github.com/rsksmart/rootstock-integration-tests/pull/103#discussion_r1786088735) comment

Vovchyk (approved):

marcos-iov (approved):

PR #106, opened 2024-10-02
Adds simple peginv1 test

jeremy-then:


lib/tests/2wp.js
```diff
@@ -105,13 +106,54 @@ const execute = (description, getRskHost) => {
...
+  
+      // Act
+  
+      const peginV1Data = [Buffer.from(createPeginV1TxData(peginV1RskRecipientAddress), 'hex')];
+  
+      const btcPeginTxHash = await sendPegin(rskTxHelper, btcTxHelper, senderRecipientInfo.btcSenderAddressInfo, Number(satoshisToBtc(peginValueInSatoshis)), peginV1Data);
+  
+      // Assert
+  
+      const expectedCountOfThisPeginUtxosInTheBridge = 1;
```
marcos-iov, line 127: maybe this could be the default value for the param in `ensurePeginIsRegistered`? What do you think?

lib/tests/2wp.js
```diff
@@ -105,13 +106,54 @@ const execute = (description, getRskHost) => {
...
+      await ensurePeginIsRegistered(rskTxHelper, btcPeginTxHash, expectedCountOfThisPeginUtxosInTheBridge);
+  
+      const expectedPeginProtocolVersion = '1';
```
marcos-iov, line 130: Same here

lib/tests/2wp.js
```diff
@@ -105,13 +106,54 @@ const execute = (description, getRskHost) => {
 
     });
 
+    it('should do a basic pegin v1 with the exact minimum value', async () => {
```
marcos-iov, line 109: Not sure if it's under your radar. But we should also have some tests that add the btc refund address to the op return payload.

Nothing will change since there won't be a refund, but we should check that the peg-in still works when we add that data.

And then some tests with invalid payloads, including invalid refund address data

lib/tests/2wp.js
jeremy-then: It is. I just wanted to express it explicitly here. I can remove it and use it like this for multiple utxos.

lib/tests/2wp.js
jeremy-then: It defaults to '0', at the `createExpectedPeginBtcEvent` level.

lib/tests/2wp.js
jeremy-then: Yes.

lib/tests/2wp.js
jeremy-then: Removed.

marcos-iov (approved):

PR #108, opened 2024-10-02
Adds new peginv1 below minimum test

jeremy-then:
Adds 'should reject a basic pegin v1 with value exactly below minimum' test.

This test sends a pegin v1 with a value exactly below minimum (the minimum pegin value minus 1 satoshis), to assert that it should be rejected with rejection reason 5 (invalid amount).

Asserts that the sender funds are lost and that the sender's derived rsk address nor the pegin v1 rsk recipient address get any funds, but the federation balance is increased by the pegin amount while the Bridge doesn't mark that pegin tx as processed.

### DIFF
--- a/lib/tests/2wp.js
+++ b/lib/tests/2wp.js
@@ -117,7 +117,7 @@ const execute = (description, getRskHost) => {
       const senderRecipientInfo = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
       const initialSenderAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, senderRecipientInfo.btcSenderAddressInfo.address);
       const peginValueInSatoshis = minimumPeginValueInSatoshis;
-      const peginV1RskRecipientAddress = await rskTxHelper.newAccountWithSeed('');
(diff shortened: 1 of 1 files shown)


lib/tests/2wp.js
```diff
@@ -192,6 +192,51 @@ const execute = (description, getRskHost) => {
...
+      const btcPeginTxHash = await sendPegin(rskTxHelper, btcTxHelper, senderRecipientInfo.btcSenderAddressInfo, satoshisToBtc(peginValueInSatoshis), peginV1Data);
+      // Funds of a pegin with value below minimum are lost. But calling triggerRelease here to ensure that nothing will be refunded.
+      await triggerRelease(rskTxHelpers, btcTxHelper);
+
+      // Assert
+
+      // The btc pegin tx is not marked as processed by the bridge
+      await assertPeginTxHashNotProcessed(btcPeginTxHash);
+
+      await assert2wpBalancesPeginRejectedBelowMinimum(initial2wpBalances, peginValueInSatoshis);
```
marcos-iov, line 222: Didn't we rename this?

jeremy-then: I needed a rebase.

marcos-iov (approved):

PR #110, opened 2024-10-03
Adds legacy pegin from multisig and below minimum should be rejected …

jeremy-then:
…and not refund test.

### DIFF
--- a/lib/tests/2wp.js
+++ b/lib/tests/2wp.js
@@ -305,7 +305,7 @@ const execute = (description, getRskHost) => {
       // To ensure no refund is made, let's try to push any possible refund pegout.
       await triggerRelease(rskTxHelpers, btcTxHelper);
 
-      await assert2wpBalancesPeginRejectedBelowMinimum(initial2wpBalances, peginValueInSatoshis);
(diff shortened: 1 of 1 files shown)


nathanieliov (approved):

lib/tests/2wp.js
```diff
@@ -273,6 +273,44 @@ const execute = (description, getRskHost) => {
...
+
+      const blockNumberBeforePegin = await rskTxHelper.getBlockNumber();
+
+      const btcPeginTxHash = await sendPegin(rskTxHelper, btcTxHelper, multisigSenderAddressInfo, Number(satoshisToBtc(peginValueInSatoshis)));
+
+      // Assert
+
+      // The btc pegin tx is not marked as processed by the bridge.
+      const isBtcTxHashAlreadyProcessed = await bridge.methods.isBtcTxHashAlreadyProcessed(btcPeginTxHash).call();
+      expect(isBtcTxHashAlreadyProcessed).to.be.false;
```
marcos-iov, line 295: We created a method for this, right?

marcos-iov (approved):

PR #117, opened 2024-10-06
Refactors PEGOUT_EVENTS in constants to include signature.

jeremy-then:
Refactors PEGOUT_EVENTS in constants.js in preparation for the new pegout tests.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -57,12 +57,30 @@ const PEGIN_UNREFUNDABLE_REASONS = {
 };
 
 const PEGOUT_EVENTS = {
-  RELEASE_REQUEST_RECEIVED: "release_request_received",
-  RELEASE_REQUEST_REJECTED: "release_request_rejected",
-  RELEASE_REQUESTED: "release_requested",
-  BATCH_PEGOUT_CREATED: "batch_pegout_created",
-  PEGOUT_TRANSACTION_CREATED: "pegout_transaction_created",
-  PEGOUT_CONFIRMED: "pegout_confirmed"
+  RELEASE_REQUEST_RECEIVED: {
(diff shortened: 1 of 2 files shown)


nathanieliov (approved):

lib/constants.js
```diff
@@ -57,12 +57,30 @@ const PEGIN_UNREFUNDABLE_REASONS = {
...
 const PEGOUT_EVENTS = {
-  RELEASE_REQUEST_RECEIVED: "release_request_received",
-  RELEASE_REQUEST_REJECTED: "release_request_rejected",
-  RELEASE_REQUESTED: "release_requested",
-  BATCH_PEGOUT_CREATED: "batch_pegout_created",
-  PEGOUT_TRANSACTION_CREATED: "pegout_transaction_created",
-  PEGOUT_CONFIRMED: "pegout_confirmed"
+  RELEASE_REQUEST_RECEIVED: {
+    name: 'release_request_received',
+    signature: '0x1a4457a4460d48b40c5280955faf8e4685fa73f0866f7d8f573bdd8e64aca5b1'
```
nathanieliov, line 62: How is this signature obtained? Do you mind adding a reference? Wdyt?

marcos-iov: @jeremy-then what do you think of using [this](https://web3js.readthedocs.io/en/v1.2.11/web3-eth-abi.html#encodeeventsignature)?

We could get the event object from precompiled-abis, and then use web3-abi dependency to encode it. Seems a bit more robust.

Or maybe precompiled-abis library could have a `getSignature` function?

No big deal anyways, just an idea. It's not something we expect to change very often.

marcos-iov (approved):

PR #120, opened 2024-10-06
Adds 'should do a pegin v1 with multiple OP_RETURN outputs but only o…

jeremy-then:
…ne for RSK and the exact minimum value' test


nathanieliov (approved):

apancorb (approved):
LGTM

marcos-iov (approved):

PR #140, opened 2024-10-22
Adds 'should reject and refund a pegout when fee per kb is above valu…

jeremy-then:
Adds 'should reject and refund a pegout when fee per kb is above value' test


lib/tests/2wp.js
```diff
@@ -777,6 +777,47 @@ const execute = (description, getRskHost) => {
 
     });
 
+    it('should reject and refund a pegout when fee per kb is above value', async () => {
+
+      // Arrange
+
+      // Create a pegin for the sender to ensure there is enough funds to pegout and because this is the natural process
+      const senderRecipientInfo = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
+      const peginValueInSatoshis = btcToSatoshis(0.5);
```
apancorb, line 786: Maybe we can have a default constant for 0.5 as well?

apancorb (approved):
LGTM

lib/tests/2wp.js
jeremy-then: I think there's no need for this. This is not the minimum pegin value or any significan value. This is just to have enough funds on the rsk address. This value could change from test to test. For example, in the future pegout test where we will need to create a lot of pegouts to assert that the Bridge split them, we will probably need to send way more funds. So, that value could change.

nathanieliov: As long as this value does not repeat in other places, it is ok to me.

lib/tests/2wp.js
```diff
@@ -777,6 +777,47 @@ const execute = (description, getRskHost) => {
...
+      const btcPeginTxHash = await sendPegin(rskTxHelper, btcTxHelper, senderRecipientInfo.btcSenderAddressInfo, satoshisToBtc(peginValueInSatoshis));
+      await ensurePeginIsRegistered(rskTxHelper, btcPeginTxHash);
+
+      const initial2wpBalances = await get2wpBalances(rskTxHelper, btcTxHelper);
+      const initialBtcRecipientAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, senderRecipientInfo.btcSenderAddressInfo.address);
+      const initialRskSenderBalanceInWeisBN = await rskTxHelper.getBalance(senderRecipientInfo.rskRecipientRskAddressInfo.address);
+      const pegoutValueInRbtc = MINIMUM_PEGOUT_AMOUNT_IN_RBTC;
+
+      // The minimum pegout value plus 1 satoshi to make the fee per kb exactly above the pegout value.
+      const newFeePerKbInSatoshis = btcToSatoshis(MINIMUM_PEGOUT_AMOUNT_IN_RBTC) + 1;
```
nathanieliov, line 796: According to the logic in [rskj](https://github.com/rsksmart/rskj/blob/8c1e3a055cb11a2738e47122fa3f96fe3bba8fc0/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L880), setting the fee per kb to the minimum peg-out + 1 is way above the peg-out value.

I think you should use `getEstimatedFeesForNextPegOutEvent` to check if the new fee set really makes the estimated fees above the sent amount.

lib/constants.js
```diff
@@ -116,6 +116,9 @@ const PEGIN_V1_RSKT_PREFIX_HEX =  '52534b54';
 
 const MINIMUM_PEGOUT_AMOUNT_IN_RBTC = 0.0025;
 
+const FEE_PER_KB_CHANGE_PK = '6a4b49312b91e203ddfb9bc2d900ebbd46fbede46a7462e770bedcb11ad405e9';
```
marcos-iov, line 119: A bit nit from my side, but whenever I see `PK` I wish I could tell if it means public or private key. And the difference is quite important

lib/rsk-utils.js
```diff
@@ -502,7 +506,25 @@ const getFedsPubKeys = async (bridge) => {
...
+const setFeePerKb = async (rskTxHelper, feePerKbInSatoshis) => {
+
+  await rskTxHelper.getClient().eth.personal.importRawKey(FEE_PER_KB_CHANGE_PK, '');
+  await rskTxHelper.getClient().eth.personal.unlockAccount(FEE_PER_KB_CHANGE_ADDR, '');
+  const bridge = getBridge(rskTxHelper.getClient());
+
+  await sendTransaction(rskTxHelper, bridge.methods.voteFeePerKbChange(feePerKbInSatoshis), FEE_PER_KB_CHANGE_ADDR);
+
+  const finalFeePerKb = await bridge.methods.getFeePerKb().call();
+  expect(Number(finalFeePerKb)).to.equal(Number(feePerKbInSatoshis));
```
marcos-iov, line 525: You could also do a `call` before sending the transaction, and assert the response code

lib/tests/2wp.js
marcos-iov: Left a similar comment in another PR. 

Not a big deal, open to discuss it. If the value won't always be the same then maybe it doesn't make sense to have a constant. But it could maybe at least be a constant within the test to not make it a magic number.

Option 2 is to calculate this value depending on what the test needs. If we are doing a peg-out for `pegoutValue=X`, then this value could be `pegoutValue * 2`

Open to hear other ideas

lib/tests/2wp.js
```diff
@@ -777,6 +777,53 @@ const execute = (description, getRskHost) => {
...
 
+    it('should reject and refund a pegout when fee per kb is above value', async () => {
+
+      // Arrange
+
+      // Create a pegin for the sender to ensure there is enough funds to pegout and because this is the natural process
+      const senderRecipientInfo = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
+      const peginValueInSatoshis = btcToSatoshis(0.5);
+      const btcPeginTxHash = await sendPegin(rskTxHelper, btcTxHelper, senderRecipientInfo.btcSenderAddressInfo, satoshisToBtc(peginValueInSatoshis));
+      await ensurePeginIsRegistered(rskTxHelper, btcPeginTxHash);
```
marcos-iov, line 788: Could this whole part maybe be moved to a method? `fundBridgeForPegout` or something of the sort. What do you think?

jeremy-then: Updated.

lib/rsk-utils.js
jeremy-then: A `call` to `voteFeePerKbChange` or `getFeePerKb`?

lib/tests/2wp.js
```diff
@@ -774,6 +774,51 @@ const execute = (description, getRskHost) => {
...
+      // Arrange
+
+      // Create a pegin for the sender to ensure there is enough funds to pegout and because this is the natural process
+      const senderRecipientInfo = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
+      await fundRskAccountThroughAPegin(rskTxHelper, btcTxHelper, senderRecipientInfo.btcSenderAddressInfo);
+
+      const initial2wpBalances = await get2wpBalances(rskTxHelper, btcTxHelper);
+      const initialBtcRecipientAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, senderRecipientInfo.btcSenderAddressInfo.address);
+      const initialRskSenderBalanceInWeisBN = await rskTxHelper.getBalance(senderRecipientInfo.rskRecipientRskAddressInfo.address);
+      const pegoutValueInSatoshis = MINIMUM_PEGOUT_AMOUNT_IN_SATOSHIS - 1;
```
marcos-iov, line 788: Is this a good idea? We want the pegout to be rejected for the fee value, I wouldn't see an invalid pegout request. That could create some confusion

lib/tests/2wp.js

lib/rsk-utils.js
jeremy-then: Updated, by calling `sendTxWithCheck`.

marcos-iov (approved):

PR #152, opened 2024-10-24
Disable rskip427 temporarily

marcos-iov:
Disabling RSKIP427 temporarily since the implementation is not compatible with the current tests. Tests for the changes implemented in RSKIP427 will be implemented as part of the ongoing refactor to the RITs

`rskj:rskip427-integration`


julia-zack (approved):

nathanieliov (approved):

jeremy-then (approved):
LGTM.

PR #156, opened 2024-10-28
Adds 'should reject and not refund a pegout when a contract is trying…

jeremy-then:
Adds 'should reject and not refund a pegout when a contract is trying to execute it' test

### DIFF
--- a/lib/contractDeployer.js
+++ b/lib/contractDeployer.js
@@ -1,0 +1,40 @@
+
+const fs = require('fs');
+const path = require('path');
+const { btcToWeis } = require('@rsksmart/btc-eth-unit-converter');
+const solUtils = require('./sol-utils');
+const TEST_RELEASE_BTC_CONTRACT = '../contracts/CallReleaseBtcContract.sol';
+const TEST_RELEASE_BTC_CONTRACT_NAME = 'CallReleaseBtcContract';
+const SOLIDITY_COMPILER_VERSION = 'v0.8.26+commit.8a97fa7a';
+const { sendFromCow } = require('./rsk-utils');
+
+const deployCallReleaseBtcContract = async (rskTxHelper) => {
+
+    const address = await rskTxHelper.getClient().eth.personal.newAccount('');
+    await sendFromCow(rskTxHelper, address, Number(btcToWeis(0.5)));
+    await rskTxHelper.getClient().eth.personal.unlockAccount(address, '');
+  
+    const fullPath = path.resolve(__dirname, TEST_RELEASE_BTC_CONTRACT);
+    const source = fs.readFileSync(fullPath).toString();
+  
+    const callReleaseBtcContract = await solUtils.compileAndDeploy(
+      SOLIDITY_COMPILER_VERSION,
+      source,
+      TEST_RELEASE_BTC_CONTRACT_NAME,
+      [],
+      rskTxHelper,
+      {
+        from: address
+      }
+    );
+
+    return {
+        creatorAddress: address,
+        callReleaseBtcContract,
+    };
+  
+};
+
+module.exports = {
+    deployCallReleaseBtcContract,
+};
--- a/lib/contractDeployer.js
+++ b/lib/contractDeployer.js
@@ -1,19 +1,18 @@
 
 const fs = require('fs');
 const path = require('path');
-const { btcToWeis } = require('@rsksmart/btc-eth-unit-converter');
 const solUtils = require('./sol-utils');
 const TEST_RELEASE_BTC_CONTRACT = '../contracts/CallReleaseBtcContract.sol';
 const TEST_RELEASE_BTC_CONTRACT_NAME = 'CallReleaseBtcContract';
 const SOLIDITY_COMPILER_VERSION = 'v0.8.26+commit.8a97fa7a';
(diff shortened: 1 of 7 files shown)


lib/contractDeployer.js
```diff
@@ -0,0 +1,40 @@
...
+const solUtils = require('./sol-utils');
+const TEST_RELEASE_BTC_CONTRACT = '../contracts/CallReleaseBtcContract.sol';
+const TEST_RELEASE_BTC_CONTRACT_NAME = 'CallReleaseBtcContract';
+const SOLIDITY_COMPILER_VERSION = 'v0.8.26+commit.8a97fa7a';
+const { sendFromCow } = require('./rsk-utils');
+
+const deployCallReleaseBtcContract = async (rskTxHelper) => {
+
+    const address = await rskTxHelper.getClient().eth.personal.newAccount('');
+    await sendFromCow(rskTxHelper, address, Number(btcToWeis(0.5)));
```
apancorb, line 14: I think we have a constant for 0.5 now? maybe I am wrong :)

apancorb (approved):
Very cool test

lib/contractDeployer.js
jeremy-then: Good catch! Updated.

nathanieliov (approved):

lib/contractDeployer.js
```diff
@@ -0,0 +1,41 @@
+
+const fs = require('fs');
+const path = require('path');
+const { btcToWeis } = require('@rsksmart/btc-eth-unit-converter');
+const solUtils = require('./sol-utils');
+const TEST_RELEASE_BTC_CONTRACT = '../contracts/CallReleaseBtcContract.sol';
+const TEST_RELEASE_BTC_CONTRACT_NAME = 'CallReleaseBtcContract';
```
marcos-iov, line 7: If this a generic contract deployer perhaps these values should be received by param then?

lib/contractDeployer.js
```diff
@@ -0,0 +1,41 @@
...
+      TEST_RELEASE_BTC_CONTRACT_NAME,
+      [],
+      rskTxHelper,
+      {
+        from: address
+      }
+    );
+
+    return {
+        creatorAddress: address,
```
marcos-iov, line 33: Is this valuable? Shouldn't we simply return the contract address?

lib/tests/2wp.js
```diff
@@ -819,6 +821,37 @@ const execute = (description, getRskHost) => {
...
+      // Assert
+
+      const contractAddressChecksummed = rskTxHelper.getClient().utils.toChecksumAddress(ensure0x(callReleaseBtcContract.options.address));
+      const expectedEvent = createExpectedReleaseRequestRejectedEvent(contractAddressChecksummed, pegoutValueInSatoshis, PEGOUT_REJECTION_REASONS.CALLER_CONTRACT);
+      const actualReleaseRequestRejectedEvent = getReleaseRequestRejectedEventFromContractCallTxReceipt(contractCallTxReceipt);
+      expect(actualReleaseRequestRejectedEvent).to.be.deep.equal(expectedEvent);
+
+      await assert2wpBalancesAfterPegoutFromContract(initial2wpBalances, pegoutValueInSatoshis);
+
+      // The rsk sender should lose the funds since there's no refund when a smart contract is trying to do a pegout
```
marcos-iov, line 848: Should we also be checking the contract address balance?

lib/contractDeployer.js
jeremy-then: The `deployCallReleaseBtcContract` function is not generic. Look at the name. The idea for this file is that if we need to deploy more specific contracts in the future, we can create a specific function for that.
What do you think? I can make it generic, no problem. Let me know.

lib/contractDeployer.js
jeremy-then: Updated. Returning the contract instance now, since it is needed.

lib/tests/2wp.js
```diff
@@ -819,6 +821,37 @@ const execute = (description, getRskHost) => {
...
 
+    it('should reject and not refund a pegout when a contract is trying to execute it', async () => {
+
+      // Arrange
+
+      const initial2wpBalances = await get2wpBalances(rskTxHelper, btcTxHelper);
+      const pegoutValueInSatoshis = MINIMUM_PEGOUT_AMOUNT_IN_SATOSHIS;
+
+      const { callReleaseBtcContract, creatorAddress } = await deployCallReleaseBtcContract(rskTxHelper);
+      const initialRskSenderBalanceInWeisBN = await rskTxHelper.getBalance(creatorAddress);
```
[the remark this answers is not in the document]
jeremy-then: Updated. Changed the logic a bit to pass the from address.

lib/tests/2wp.js
jeremy-then: Updated.

lib/contractDeployer.js
marcos-iov: Ok, thanks. I know there are contracts being deployed in other tests, perhaps we could eventually find a way to somehow concentrate that logic here. But out of the scope for now, I'll just create a ticket to analyze further

marcos-iov (approved):

PR #158, opened 2024-10-28
Add pegout with weis rounded down to satoshis test

jeremy-then:
Add pegout with weis rounded down to satoshis test

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -124,9 +124,6 @@ const FEE_PER_KB_RESPONSE_CODES = {
   SUCCESSFUL_VOTE: 1,
 };
 
-const FEE_PER_KB_CHANGE_PK = '6a4b49312b91e203ddfb9bc2d900ebbd46fbede46a7462e770bedcb11ad405e9';
-const FEE_PER_KB_CHANGE_ADDR = '53f8f6dabd612b6137215ddd7758bb5cdd638922';
-
 module.exports = {
   KEY_TYPE_BTC,
   KEY_TYPE_RSK,
--- a/lib/tests/2wp.js
+++ b/lib/tests/2wp.js
@@ -856,11 +856,8 @@ const execute = (description, getRskHost) => {
 
       // Arrange
 
-      // Create a pegin for the sender to ensure there is enough funds to pegout and because this is the natural process
       const senderRecipientInfo = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
-      const peginValueInSatoshis = btcToSatoshis(0.5);
-      const btcPeginTxHash = await sendPegin(rskTxHelper, btcTxHelper, senderRecipientInfo.btcSenderAddressInfo, satoshisToBtc(peginValueInSatoshis));
-      await ensurePeginIsRegistered(rskTxHelper, btcPeginTxHash);
+      await fundRskAccountThroughAPegin(rskTxHelper, btcTxHelper, senderRecipientInfo.btcSenderAddressInfo);
 
       const initial2wpBalances = await get2wpBalances(rskTxHelper, btcTxHelper);
       const initialSenderAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, senderRecipientInfo.btcSenderAddressInfo.address);


apancorb (approved):
LGTM

lib/2wp-utils.js
```diff
@@ -64,7 +64,7 @@ const sendTxToBridge = async (rskTxHelper, amountInRbtc, rskFromAddress, mine =
     const txPromise = rskTxHelper.sendTransaction({
       from: rskFromAddress,
       to: BRIDGE_ADDRESS,
-      value: Number(btcEthUnitConverter.btcToWeis(amountInRbtc)),
+      value: Number(btcEthUnitConverter.ethToWeis(amountInRbtc)),
```
nathanieliov, line 67: Why `ethToWeis` if you pass the amount in `RBTC`? Am I missing something?

lib/tests/2wp.js
```diff
@@ -863,6 +863,56 @@ const execute = (description, getRskHost) => {
...
+      // Checking all the pegout events are emitted and in order
+      const blockNumberAfterPegoutRelease = await rskTxHelper.getBlockNumber();
+      const pegoutsEvents = await getPegoutEventsInBlockRange(rskTxHelper, pegoutTransaction.blockNumber, blockNumberAfterPegoutRelease);
+      await assertSuccessfulPegoutEventsAreEmitted(pegoutsEvents, pegoutTransaction.transactionHash, senderRecipientInfo, expectedPegoutValueInRbtc, bridgeStateAfterPegoutCreation);
+
+      await assert2wpBalanceAfterSuccessfulPegoutWithLargeWeis(initial2wpBalances, pegoutValueInRbtc, btcToSatoshis(expectedPegoutValueInRbtc));
+
+      // Assert that the sender address balance is increased by the actual pegout value
+      const finalSenderAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, senderRecipientInfo.btcSenderAddressInfo.address);
+      const releaseBtcEvent = pegoutsEvents[pegoutsEvents.length - 1];
```
nathanieliov, line 909: I think we can abstract this logic of getting the release event using the last index into a method, just to give a bit of context on what this means `pegoutsEvents.length - 1`. Wdyt?

lib/2wp-utils.js
jeremy-then: RBTC is the same as ETH. Anyways, this was updated with a different logic.

lib/tests/2wp.js
jeremy-then: Well, the `release_btc` event is the last pegout related event emitted. I'm using indices in the tests to access the vents because this way we also check that the events are being emitted in the correct order.
I'm not using things like `Array::find`, etc., because of what I explained above and also because that would be inefficient.

nathanieliov (approved):

lib/constants.js
```diff
@@ -124,6 +124,9 @@ const FEE_PER_KB_RESPONSE_CODES = {
   SUCCESSFUL_VOTE: 1,
 };
 
+const FEE_PER_KB_CHANGE_PK = '6a4b49312b91e203ddfb9bc2d900ebbd46fbede46a7462e770bedcb11ad405e9';
+const FEE_PER_KB_CHANGE_ADDR = '53f8f6dabd612b6137215ddd7758bb5cdd638922';
+
```
marcos-iov, line 129: Wasn't this previously added?

lib/tests/2wp.js
```diff
@@ -852,6 +852,56 @@ const execute = (description, getRskHost) => {
...
 
+    it('should do a pegout and round down the weis to satoshis as expected', async () => {
+
+      // Arrange
+
+      // Create a pegin for the sender to ensure there is enough funds to pegout and because this is the natural process
+      const senderRecipientInfo = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
+      const peginValueInSatoshis = btcToSatoshis(0.5);
+      const btcPeginTxHash = await sendPegin(rskTxHelper, btcTxHelper, senderRecipientInfo.btcSenderAddressInfo, satoshisToBtc(peginValueInSatoshis));
+      await ensurePeginIsRegistered(rskTxHelper, btcPeginTxHash);
```
marcos-iov, line 863: Isn't there a new util function for this?

lib/tests/2wp.js
```diff
@@ -852,6 +852,56 @@ const execute = (description, getRskHost) => {
...
+
+      const initial2wpBalances = await get2wpBalances(rskTxHelper, btcTxHelper);
+      const initialSenderAddressBalanceInSatoshis = await getBtcAddressBalanceInSatoshis(btcTxHelper, senderRecipientInfo.btcSenderAddressInfo.address);
+      const pegoutValueInWeisBN = new BN('41234567891234560'); // 0.04123456789123456 in RBTC
+      const expectedPegoutValueInSatoshis = 4123456; // 0.04123456 in BTC
```
marcos-iov, line 868: Can't we use the unit converter library to somehow make this clearer?

jeremy-then: It turns out, we cant. The unit converter rounds differently than the Bridge. The unit converter would yield `4123457`, and the Bridge `4123456`. So, I had to hardcode it.

lib/tests/2wp.js

lib/constants.js
jeremy-then: Updated.

lib/tests/2wp.js
marcos-iov: Should we update the unit converter library then?

PR #159, opened 2024-10-29
Fixing the use of RIT branch instead of hardcoded

fmacleal:
Adding the configuration of branches from PR and configurable and removing the hardcoded.

`rskj:snap-v4-final`


nathanieliov (approved):

jeremy-then (approved):
LGTM.

PR #163, opened 2024-10-31
Adds 'should create multiple pegouts with mixed values same and above…

jeremy-then:
Adds 'should create multiple pegouts with mixed values same and above minimum' test

### DIFF
--- a/lib/2wp-utils.js
+++ b/lib/2wp-utils.js
@@ -59,17 +59,19 @@ const assertRefundUtxosSameAsPeginUtxos = async(rskTxHelper, btcTxHelper, peginT
  * @param {BN} amountInWeisBN 
  * @param {string} rskFromAddress 
  * @param {boolean} mine If true, mines 1 block after sending the transaction. If false, it will not mine the tx and will return undefined. Defaults to true.
- * @returns {web3.eth.TransactionReceipt | undefined}
+ * @returns {Promise<web3.eth.TransactionReceipt | txPromise>} the rsk tx receipt if `mine` is true, otherwise the tx promise.
  */
 const sendTxToBridge = async (rskTxHelper, amountInWeisBN, rskFromAddress, mine = true) => {
+
     const txPromise = rskTxHelper.sendTransaction({
       from: rskFromAddress,
       to: BRIDGE_ADDRESS,
       value: amountInWeisBN,
       gasPrice: TO_BRIDGE_GAS_PRICE,
     });
+
     if(!mine) {
-        return;
+        return txPromise;
     }
 
     // Wait for the rsk tx to be in the rsk mempool before mining
@@ -78,6 +80,7 @@ const sendTxToBridge = async (rskTxHelper, amountInWeisBN, rskFromAddress, mine 
     await mineAndSync(getRskTransactionHelpers());
     const result = await txPromise;
     return result;
+    
 };
 
 /**
--- a/lib/btc-utils.js
+++ b/lib/btc-utils.js
@@ -187,6 +187,10 @@ const waitForBitcoinMempoolToGetTxs = async (btcTxHelper, maxAttempts = 3, check
     return Number(btcToSatoshis(await btcTxHelper.getAddressBalance(btcAddress)));
   };
 
+  const base58AddressToHexString = (base58Address) => {
+    return bitcoinJs.address.fromBase58Check(base58Address).hash.toString('hex');
+  };
+
   module.exports = {
     publicKeyToCompressed,
     fundAddressAndGetData,
@@ -194,4 +198,5 @@ const waitForBitcoinMempoolToGetTxs = async (btcTxHelper, maxAttempts = 3, check
     waitForBitcoinTxToBeInMempool,
     waitForBitcoinMempoolToGetTxs,
     getBtcAddressBalanceInSatoshis,
+    base58AddressToHexString,
   };
--- a/lib/btc-utils.js
+++ b/lib/btc-utils.js
@@ -181,22 +181,27 @@ const waitForBitcoinMempoolToGetTxs = async (btcTxHelper, maxAttempts = 3, check
     logger.debug(`[${waitForBitcoinMempoolToGetTxs.name}] The final bitcoin mempool size is ${finalBitcoinMempoolSize}, after ${attempts} attempts. Difference with initial mempool size: ${finalBitcoinMempoolSize - initialBitcoinMempoolSize}.`);
   
     return bitcoinMempoolHasTx;
-  }
+}
 
-  const getBtcAddressBalanceInSatoshis = async (btcTxHelper, btcAddress) => {
-    return Number(btcToSatoshis(await btcTxHelper.getAddressBalance(btcAddress)));
-  };
+const getBtcAddressBalanceInSatoshis = async (btcTxHelper, btcAddress) => {
+  return Number(btcToSatoshis(await btcTxHelper.getAddressBalance(btcAddress)));
+};
 
-  const base58AddressToHexString = (base58Address) => {
-    return bitcoinJs.address.fromBase58Check(base58Address).hash.toString('hex');
-  };
+/**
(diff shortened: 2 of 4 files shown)


apancorb (approved):
LGTM, minor comments

lib/2wp-utils.js
```diff
@@ -59,17 +59,19 @@ const assertRefundUtxosSameAsPeginUtxos = async(rskTxHelper, btcTxHelper, peginT
...
 const sendTxToBridge = async (rskTxHelper, amountInWeisBN, rskFromAddress, mine = true) => {
+
     const txPromise = rskTxHelper.sendTransaction({
       from: rskFromAddress,
       to: BRIDGE_ADDRESS,
       value: amountInWeisBN,
       gasPrice: TO_BRIDGE_GAS_PRICE,
     });
+
     if(!mine) {
```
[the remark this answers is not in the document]
jeremy-then: I just noticed there is a mixed of `if (` an `if(` in the test. I always prefer `if(`. But what should we always use? @nathanieliov , @marcos-iov , @julia-zack .

lib/tests/2wp.js
```diff
@@ -902,6 +910,130 @@ const execute = (description, getRskHost) => {
...
+        bridgeStateAfterPegoutCreation = await getBridgeState(rskTxHelper.getClient());
+      };
+
+      const callbacks = {
+        pegoutCreatedCallback
+      };
+
+      await triggerRelease(rskTxHelpers, btcTxHelper, callbacks);
+
+      // Checking all the pegout events are emitted and in order
```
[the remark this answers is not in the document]
jeremy-then: They all get mined in the same block, otherwise, the test would fail.
Since they can get mined in different order, I'm forced to use `pegoutsEvents.find` index of indices and `getAddressUtxo` to find the corresponding utxos.

lib/rsk-utils.js
```diff
@@ -529,6 +529,30 @@ const setFeePerKb = async (rskTxHelper, feePerKbInSatoshis) => {
...
+ * @param {RskTransactionHelper} rskTxHelper 
+ * @param {number} atLeastExpectedCount 
+ * @param {number} waitTimeInMilliseconds defaults to 500 milliseconds
+ * @param {number} maxAttempts defaults to 20 attempts
+ * @returns {Promise<void>}
+ * @throws {Error} if the rsk mempool doesn't get at least `atLeastExpectedCount` txs after `maxAttempts` attempts
+ */
+const waitForRskMempoolToGetAtLeastThisManyTxs = async (rskTxHelper, atLeastExpectedCount, waitTimeInMilliseconds = 500, maxAttempts = 20) => {
+  let attempts = 1;
+  while(attempts <= maxAttempts) {
```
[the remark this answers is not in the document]
jeremy-then: It would then look like the following if we wanted to have access to the `attempts` variable:

```js
const waitForRskMempoolToGetAtLeastThisManyTxs = async (rskTxHelper, atLeastExpectedCount, waitTimeInMilliseconds = 500, maxAttempts = 20) => {
  let attempts = 1;
  for(; attempts <= maxAttempts; attempts++) {
    const rskMempoolTxs = await getRskMempoolTransactionHashes(rskTxHelper);
    logger.debug(`[${waitForRskMempoolToGetAtLeastThisManyTxs.name}] rsk mempool txs: ${rskMempoolTxs.length}`);
    if(rskMempoolTxs.length >= atLeastExpectedCount) {
      logger.debug(`[${waitForRskMempoolToGetAtLeastThisManyTxs.name}] rsk mempool has ${rskMempoolTxs.length} (at least expected was ${atLeastExpectedCount}) txs after ${attempts} attempts`);
      return;
    }
    await wait(waitTimeInMilliseconds);
  }

  throw new Error(`[${waitForRskMempoolToGetAtLeastThisManyTxs.name}] rsk mempool didn't get at least ${atLeastExpectedCount} txs after ${maxAttempts} attempts`);
};
```

By declaring it in the `for` like `for(let attemptt...`, we wouldn't have access it to it below the `for`. And with it declared above the `for`, then there's not really much difference than with the `while`, in fact, looks weirder this part like that.

lib/tests/2wp.js
```diff
@@ -902,6 +910,130 @@ const execute = (description, getRskHost) => {
...
+      // Arrange
+      
+      const senderRecipientInfo1 = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
+      await fundRskAccountThroughAPegin(rskTxHelper, btcTxHelper, senderRecipientInfo1.btcSenderAddressInfo);
+      
+      const senderRecipientInfo2 = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
+      await fundRskAccountThroughAPegin(rskTxHelper, btcTxHelper, senderRecipientInfo2.btcSenderAddressInfo);
+
+      const senderRecipientInfo3 = await createSenderRecipientInfo(rskTxHelper, btcTxHelper);
+      await fundRskAccountThroughAPegin(rskTxHelper, btcTxHelper, senderRecipientInfo3.btcSenderAddressInfo);
```
nathanieliov, line 924: I think it would be better to add all the recipients into an array, then iterate that array for each operation and the same for the assertions.

lib/2wp-utils.js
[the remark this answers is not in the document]
nathanieliov: @apancorb 's suggestion, as it seems to be the rule of our formatting settings. I formatted it in my local and the result was @apancorb suggestion.

lib/rsk-utils.js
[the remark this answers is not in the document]
nathanieliov: I would vote to keep as it is now.

nathanieliov (approved):

lib/rsk-utils.js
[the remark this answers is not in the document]
apancorb: I see, looks good to me

lib/tests/2wp.js
```diff
@@ -905,6 +914,130 @@ const execute = (description, getRskHost) => {
...
+      const pegoutTransaction3 = await pegoutTransaction3Promise;
+
+      // Assert
+
+      let bridgeStateAfterPegoutCreation;
+      
+      // Callback to get the bridge state after the pegout is created
+      const pegoutCreatedCallback = async () => {
+        bridgeStateAfterPegoutCreation = await getBridgeState(rskTxHelper.getClient());
+      };
```
marcos-iov, line 970: I might be missing something, what's the value of this callback? It will simply get the Bridge state and do nothing with it

lib/tests/2wp.js
```diff
@@ -905,6 +914,130 @@ const execute = (description, getRskHost) => {
...
+
+      // Now our 3 pegout transaction requests will get mined together.
+      await rskTxHelper.mine();
+      
+      // After mining, we can get the transaction receipts.
+      const pegoutTransaction1 = await pegoutTransaction1Promise;
+      const pegoutTransaction2 = await pegoutTransaction2Promise;
+      const pegoutTransaction3 = await pegoutTransaction3Promise;
+
+      // Assert
```
marcos-iov, line 963: Should we check the Bridge state at this point? To assert there are 3 pegout requests created

jeremy-then: Good idea.

jeremy-then: Updated.

lib/btc-utils.js
```diff
@@ -187,11 +187,16 @@ const waitForBitcoinMempoolToGetTxs = async (btcTxHelper, maxAttempts = 3, check
     return Number(btcToSatoshis(await btcTxHelper.getAddressBalance(btcAddress)));
   };
 
+  const base58AddressToHexString = (base58Address) => {
```
marcos-iov, line 190: `toHexString` doesn't say much. Maybe we should call this `toHash160`? If that is the case, not sure what the hex string is representing here

jeremy-then: Updated.

marcos-iov (approved):

PR #200, opened 2025-01-14
Settup and add initial svp tests.

jeremy-then:


lib/rsk-utils.js
```diff
@@ -477,7 +480,7 @@ const findBridgeTransactionsInThisBlock = async (web3Client, blockHashOrBlockNum
   // const bridgeTransactionParser = new BridgeTransactionParser(web3Client);
   // return bridgeTransactionParser.getBridgeTransactionsInThisBlock(blockHashOrBlockNumber);
 
-  return bridgeTxParser.getBridgeTransactionsInThisBlock(blockHashOrBlockNumber);
+  return await bridgeTxParser.getBridgeTransactionsInThisBlock(blockHashOrBlockNumber);
```
nathanieliov, line 483: Should we remove the `async` from the definition of the method?

tests/00_00_04-change-federation.js
```diff
@@ -236,6 +251,111 @@ describe('Change federation', async function() {
 
     });
 
+    it('should create the proposed federation', async () => {
+
+        // Act
+
+        const proposedFederationInfo = await getProposedFederationInfo(bridge);
```
nathanieliov, line 258: I think this is not related to the act of creating the proposed federation, therefore I would put this in the assert section.

lib/federation-utils.js
```diff
@@ -23,7 +23,81 @@ var compareFederateKeys = (publicKeysA, publicKeysB) => {
   return comparison;
 };
 
+const getActiveFederationKeys = async (bridge) => {
```
julia-zack, line 26:
```suggestion
const getActiveFederationPublicKeys = async (bridge) => {
```

just to be consistent with the other name

tests/00_00_04-change-federation.js
```diff
@@ -236,6 +251,111 @@ describe('Change federation', async function() {
...
+        // Assert
+
+        const expectedProposedFederationSize = newFederationPublicKeys.length;
+        expect(proposedFederationInfo.proposedFederationSize).to.be.equal(expectedProposedFederationSize, 'The proposed federation size should be the expected one.');
+
+        expect(proposedFederationInfo.proposedFederationAddress).to.be.equal(expectedNewFederationAddress, 'The proposed federation address should be the expected one.');
+
+        expect(proposedFederationInfo.proposedFederationCreationBlockNumber).to.be.equal(commitFederationCreationBlockNumber, 'The proposed federation creation block number should be the expected one.');
+        
+        const proposedFederationMembers = await getProposedFederationPublicKeys(bridge);
```
julia-zack, line 269:
```suggestion
        const proposedFederationPublicKeys = await getProposedFederationPublicKeys(bridge);
```

tests/00_00_04-change-federation.js
```diff
@@ -236,6 +251,111 @@ describe('Change federation', async function() {
...
+
+        expect(proposedFederationMembers).to.be.deep.equal(newFederationPublicKeys, 'The proposed federation members should be the expected ones.');
```
julia-zack, line 271:
```suggestion
        expect(proposedFederationMembers).to.be.deep.equal(newFederationPublicKeys, 'The proposed federation public keys should be the expected ones.');
```

tests/00_00_04-change-federation.js
```diff
@@ -236,6 +251,111 @@ describe('Change federation', async function() {
...
+
+        // The output addresses should be in the expected order.
+        const proposedFederationOutput = btcTransaction.outs[0];
+        const flyoverOutput = btcTransaction.outs[1];
+        const activeFederationOutput = btcTransaction.outs[2];
+
+        const actualProposedFederationAddress = bitcoinJsLib.address.fromOutputScript(proposedFederationOutput.script, btcTxHelper.btcConfig.network);
+        expect(actualProposedFederationAddress).to.be.equal(proposedFederationAddress, 'The proposed federation address in the SVP fund transaction should be the first output.');
+
+        const actualFlyoverAddress = bitcoinJsLib.address.fromOutputScript(flyoverOutput.script, btcTxHelper.btcConfig.network);
```
julia-zack, line 327:
```suggestion
        const actualProposedFederationFlyoverAddress = bitcoinJsLib.address.fromOutputScript(flyoverOutput.script, btcTxHelper.btcConfig.network);
```

tests/00_00_04-change-federation.js
```diff
@@ -236,6 +251,111 @@ describe('Change federation', async function() {
...
+
+        const btcTransaction = bitcoinJsLib.Transaction.fromHex(rawSvpBtcTransaction);
+
+        expect(btcTransaction.outs.length).to.be.equal(3, 'The SVP fund transaction should have 3 outputs.');
+
+        const proposedFederationAddress = await bridge.methods.getProposedFederationAddress().call();
+
+        // The output addresses should be in the expected order.
+        const proposedFederationOutput = btcTransaction.outs[0];
+        const flyoverOutput = btcTransaction.outs[1];
```
julia-zack, line 321:
```suggestion
        const proposedFederationFlyoverOutput = btcTransaction.outs[1];
```

tests/00_00_04-change-federation.js
```diff
@@ -284,3 +404,59 @@ describe('Change federation', async function() {
...
+
+    const svpSpendTxHashUnsigned = await rskTxHelper.getClient().rsk.getStorageBytesAt(BRIDGE_ADDRESS, svpSpendTxHashUnsignedStorageIndex);
+    expect(svpSpendTxHashUnsigned).to.be.equal('0x0', 'The SVP spend tx hash unsigned storage value should be empty.');
+
+    const svpSpendTxWaitingForSignatures = await rskTxHelper.getClient().rsk.getStorageBytesAt(BRIDGE_ADDRESS, svpSpendTxWaitingForSignaturesStorageIndex);
+    expect(svpSpendTxWaitingForSignatures).to.be.equal('0x0', 'The SVP spend tx waiting for signatures storage value should be empty.');
+
+};
+
+const assertPegoutTrasactionCreatedOutpointValues = async (initialBridgeState, finalBridgeState, svpBtcTransaction, pegoutTransactionCreatedEvent) => {
```
julia-zack, line 441:
```suggestion
const assertPegoutTransactionCreatedOutpointValues = async (initialBridgeState, finalBridgeState, svpBtcTransaction, pegoutTransactionCreatedEvent) => {
```

just a typo

lib/rsk-utils.js
jeremy-then: Why? We are using `await` now, so we need it.

tests/00_00_04-change-federation.js
```diff
@@ -236,6 +251,111 @@ describe('Change federation', async function() {
...
+        const pegoutCreationBlockNumber = Number(svpPegoutWaitingForConfirmations.pegoutCreationBlockNumber);
+
+        const expectedPegoutCreationBlockNumber = commitFederationCreationBlockNumber + 1;
+        expect(pegoutCreationBlockNumber).to.be.equal(expectedPegoutCreationBlockNumber, 'The svp fund tx pegout creation block number should be the block that contains the first updateCollections call right after the commitFederation call.');
+
+        const rawSvpBtcTransaction = svpPegoutWaitingForConfirmations.btcRawTx;
+
+        const btcTransaction = bitcoinJsLib.Transaction.fromHex(rawSvpBtcTransaction);
+
+        expect(btcTransaction.outs.length).to.be.equal(3, 'The SVP fund transaction should have 3 outputs.');
```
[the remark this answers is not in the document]
jeremy-then: Based on the current setup, it at least in regtest, it will always have 3 outputs.
The minimum pegin value in regtest is 0.5 btc, so, there will not be any utxo below that amount. And this amount is more than enough for the svp process, since the output to the proposed fed will be 0.005 btc, the output to the flyover address will be 0.005, so at least for the current setup that we have, there will always be a change.

tests/00_00_04-change-federation.js
jeremy-then: Updated.

lib/federation-utils.js
jeremy-then: Updated.

tests/00_00_04-change-federation.js
jeremy-then: Updated.

tests/00_00_04-change-federation.js
jeremy-then: Updated.

tests/00_00_04-change-federation.js
jeremy-then: Updated.

tests/00_00_04-change-federation.js
jeremy-then: Updated.

tests/00_00_04-change-federation.js
jeremy-then: Updated.

tests/00_00_04-change-federation.js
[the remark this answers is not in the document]
julia-zack: good, thanks

tests/00_00_04-change-federation.js
julia-zack: sure? i don't see the change

tests/00_00_04-change-federation.js
```diff
@@ -236,6 +251,107 @@ describe('Change federation', async function() {
...
+
+        expect(proposedFederationInfo.proposedFederationCreationBlockNumber).to.be.equal(commitFederationCreationBlockNumber, 'The proposed federation creation block number should be the expected one.');
+        
+        const proposedFederationMembers = await getProposedFederationPublicKeys(bridge);
+
+        expect(proposedFederationMembers).to.be.deep.equal(newFederationPublicKeys, 'The proposed federation public keys should be the expected ones.');
+
+    });
+
+    it('should not have created svp transaction and there should not be any SVP values in storage yet', async () => {
```
julia-zack, line 271:
```suggestion
    it('should not have created svp fund transaction and there should not be any SVP values in storage yet', async () => {
```

jeremy-then: Updated.

lib/federation-utils.js
```diff
@@ -23,7 +23,81 @@ var compareFederateKeys = (publicKeysA, publicKeysB) => {
   return comparison;
 };
 
+const getActiveFederationPublicKeys = async (bridge) => {
```
marcos-iov, line 26: What do you think of using [powpeg-details](https://github.com/rsksmart/powpeg-details) library?

Not now, but if you think it's a good idea we can add it in a separate PR

lib/utils.js
```diff
@@ -160,6 +161,37 @@ function splitStringIntoChunks(inputString, chunkSize) {
...
+  return result;
+};
+
+const decodeRlp = (rlpEncoded) => {
+  const uint8ArrayDecoded = RLP.decode(rlpEncoded);
+  const bytesStr = bytesToHexString(uint8ArrayDecoded);
+  return bytesStr;
+};
+
+const getBridgeStorageValueDecodedHexString = (bridgeStorageValueEncodedAsRlp, append0xPrefix = true) => {
```
marcos-iov, line 182: I think we implemented this in one of our libraries didn't we? Might be a good idea anyway to have maybe a specialized library that reads values from the Bridge storage

package.json
```diff
@@ -13,7 +13,7 @@
   "license": "GPL3",
   "dependencies": {
     "@rsksmart/bridge-state-data-parser": "^1.2.0",
-    "@rsksmart/bridge-transaction-parser": "^1.1.4",
+   "@rsksmart/bridge-transaction-parser": "github:rsksmart/bridge-transaction-parser#fbdeb686fc438f530d3a5f01bce062fd455be01f",
```
marcos-iov, line 16: Why this? Do we need a new release for `bridge-transaction-parser`?

package.json
```diff
@@ -34,7 +34,7 @@
     "mocha-junit-reporter": "^2.2.1",
     "pegin-address-verificator": "git+https://git@github.com/rsksmart/pegin-address-verifier#v0.4.0",
     "peglib": "git+https://github.com/rsksmart/rsk-peglib#v1.4.15",
-    "@rsksmart/rsk-precompiled-abis": "^6.0.0-ARROWHEAD",
+    "@rsksmart/rsk-precompiled-abis": "github:rsksmart/precompiled-abis#841aa8f0caab0356ae510708bfb2fceae9144790",
```
marcos-iov, line 37: same here

test.js
```diff
@@ -56,7 +56,7 @@ global.Runners = {
       fingerroot500: 1,
       arrowhead600: 1,
       arrowhead631: 1,
-      lovell700: -1,
+      lovell700: 1,
```
marcos-iov, line 59: I thought we planned on staring with Lovell inactive. Then test pre and post Lovell

marcos-iov (commented):
looking good

package.json
jeremy-then: Yes. We need to update the `@rsksmart/rsk-precompiled-abis` dependency in the  `bridge-transaction-parser` library. Check: https://github.com/rsksmart/bridge-transaction-parser/compare/master...use-lovel-precompiled-abis-commit

lib/federation-utils.js
jeremy-then: It only works for the active federation.

PR #205, opened 2025-01-21
Adds register svp spend tx test

jeremy-then:


tests/00_00_04-change-federation.js
```diff
@@ -382,6 +385,29 @@ describe('Change federation', async function() {
...
+        const releaseBtcTransaction = bitcoinJsLib.Transaction.fromHex(removePrefix0x(releaseBtcEvent.arguments.btcRawTransaction));
+
+        // Mining the SVP Spend transaction in bitcoin to register it in the Bridge
+        await waitForBitcoinTxToBeInMempool(btcTxHelper, releaseBtcTransaction.getId());
+        await btcTxHelper.mine(BTC_TO_RSK_MINIMUM_CONFIRMATIONS);
+        await rskUtils.waitAndUpdateBridge(rskTxHelper);
+
+        await assertProposedFederationIsNotInStorage(bridge);
+
+        await assertSvpValuesNotPresentInStorage(rskTxHelper);
```
julia-zack, line 407: i think we are missing some assertions:
- `assertNextFederationCreationBHIsSetAsExpected`
(we should save the proposedFederation.creationBN before registering the spend tx and assert they match)
- assertUtxosReferenceMovedFromNewToOld
- assertLasRetiredP2SHScriptIsSetAsExpected
- (same here, we should save the proposedFederation.P2SHScript before registering the spend tx and assert they match)

PR #208, opened 2025-01-24
Fixes 2wp tests due to lovell changes.

jeremy-then:


lib/tests/2wp.js
```diff
@@ -1197,30 +1187,33 @@ const assertReleaseBtcEvent = (releaseBtcEvent, releaseRequestedEvent) => {
   expect(releaseBtcEvent.arguments.releaseRskTxHash).to.be.equal(releaseRequestedEvent.arguments.rskTxHash);
 };
 
-const assertSuccessfulPegoutEventsAreEmitted = async (pegoutsEvents, pegoutRequestReceivedTransactionHash, senderRecipientInfo, pegoutValueInSatoshis, bridgeStateAfterPegoutCreation) => {
-
+const assertSuccessfulPegoutEventsAreEmitted = async (pegoutsEvents, pegoutRequestReceivedTransactionHash, senderRecipientInfo, pegoutValueInWeisBN, pegoutValueInSatoshis, bridgeStateBeforePegoutCreation, bridgeStateAfterPegoutCreation) => {
```
julia-zack, line 1190: i think is weird to receive both `pegoutValueInWeis` and `pegoutValueInSatoshis`. We could just receive the satoshis value and do the conversion to weis inside the method, wdyt?

lib/tests/2wp.js
```diff
@@ -997,35 +988,40 @@ const execute = (description, getRskHost) => {
...
       const pegoutWaitingForConfirmationWhenPegoutWasCreated = bridgeStateAfterPegoutCreation.pegoutsWaitingForConfirmations[0];
       const btcTransaction = bitcoinJsLib.Transaction.fromHex(pegoutWaitingForConfirmationWhenPegoutWasCreated.btcRawTx);
       const pegoutCreationRskTransactionHash = ensure0x(pegoutWaitingForConfirmationWhenPegoutWasCreated.rskTxHash.padStart(64, '0'));
 
-      // Release requested events
       const releaseRequestedEvent = pegoutsEvents[3];
       assertReleaseRequestedEvent(releaseRequestedEvent, pegoutCreationRskTransactionHash, btcTransaction.getId(), totalPegoutValueInSatoshis);
 
-      const batchPegoutCreatedEvent = pegoutsEvents[4];
+      // assertPegoutTransactionCreatedEvent()
```
julia-zack, line 1010: remove?

julia-zack (approved):

lib/tests/2wp.js
jeremy-then: Yep, I agree, but none of the conversion tools round it like the Bridge does. I haven't been able to convert the weis values to satoshis like the Bridge, so, I felt forced to hard code it here.

julia-zack: but I mean, you are passing both `satoshisToWeis(pegoutValueInSatoshis)` and `pegoutValueInSatoshis`.
Can't you just do the `satoshisToWeis(pegoutValueInSatoshis)` conversion inside the method? Maybe im missing something

julia-zack: oh okey, i see. 
what if we simplify the [expectedPegoutValueInSatoshi](https://github.com/rsksmart/rootstock-integration-tests/pull/208/files#diff-530dff31017eb3e951f3872538c0ad19594ae9ce04870a05b9f8029fc16225aeR879)? Is the same?

julia-zack (approved):

PR #217, opened 2025-02-04
Refactor logic to start federators to dynamically start and stop mult…

jeremy-then:
`rskj:run-rit`
`fed:LOVELL-7.0.0-rc`

We need to be able to run the federation change tests multiple times. We also need to be able to replace all the powpeg nodes with new one (create a new federation that doesn't contain any of the previous members).
For this, we need to be able to dynamically start and stop federate nodes for a given previous federation after a federation change.

To achieve this, some major refactor needs to be done. In this pr, I present a solution.

At the moment, 2 federations will run in the RITs: the genesis federation, and the second federation.

All `regx.key` files in `node-configs` directory were renamed to `fedx.key` and moved to the new directory `node-configs/genesis-federation` and a new directory called `second-federation` was created inside `node-configs` to hold the `fedx.key` files for the second federation.

All `node-keys/regx-v2-key.json` files were deleted since they are not being used now. We will take care of that when we implement the tcpsigner changes in the future.

`config/pipeline.conf` file was deleted. We are not using jenkins anymore.

The `regtest.js` default config file was renamed `regtest-all-keyfiles.js` and included in the git history, since now it doesn't use absolute paths, but environment variables instead, so it can now be included in the git history to easy the setup and usage of the rits. Also, the structure of that config object was changed, to include an array of federations (now only containing the genesis and second federation). And the `peer` configs were hardcoded. 

All the `config/regtest.js...` sample files were removed. They were outdated.

The `configure.sh` file was updated to point to the new locations of the `fedx.key` files.

New logback .xml files were created for the second federation. The old .xml files were moved to `logbacks/genesis-federation` directory.

The federation change tests were moved from tests/00_00_04-change-federation.js  to `lib/tests/change-federation.js` to be able to execute them multiple times. They are being executed from `tests/00_00_05-change-federation.js`. And ready to be executed from anywhere else, anytime.

A new `federationId` property was added to the config feds to be able to filter them, start them, stop them, etc.

Failing tests due to Lovell activation were fixed in this pr, to ensure this pr can fully run.

The SVP tests are not included yet. They will be included in new prs. There are open prs with the SVP changes, but they will be recreated to avoid large merge conflicts since this pr includes some changes already in other prs.


lib/assertions/fed.js
```diff
@@ -18,7 +18,7 @@ var assertKeyControl = (federates) => (begin, end, amountToTransferInWeis) => {
     var federateAddresses = federates
       .slice(begin, end)
       .map((federate) => rskClient.rsk.utils.publicKeyToAddress(
-          bitcoin.keys.publicKeyToUncompressed(federate.publicKeys.rsk)
+          bitcoin.keys.publicKeyToUncompressed(removePrefix0x(federate.publicKeys.rsk))
```
jeremy-then, line 21: Now we include the `0x` for the public keys in the config files. But this is expecting it to not have the `0x` prefix.

lib/bridge-provider.js
```diff
@@ -1,22 +1,13 @@
 const precompiledArrowhead600 = require("@rsksmart/rsk-precompiled-abis");
 
-let bridgeCache;
```
jeremy-then, line 3: After changing the federation and stopping previous federators, we don't want to cache the bridge contract instance to avoid issues.

lib/tests/activate-fork
```diff
@@ -0,0 +1,29 @@
+const { expect } = require('chai');
```
jeremy-then, line 1: This code is in master. I had removed it recently in other prs in the integration branch but we need it back now.

test.js
```diff
@@ -42,26 +41,65 @@ const bookkeepingConfigurations = {
...
+      fingerroot500: createForkObject('fingerroot500', 1),
+      arrowhead600: createForkObject('arrowhead600', 1),
+      arrowhead631: createForkObject('arrowhead631', 1),
+      lovell700: createForkObject('lovell700', 1100),
     },
     additionalFederationAddresses: []
   }
 };
 
+Runners.config = config;
```
jeremy-then, line 101: Making the configuration (regtest.js or regtest-all-keyfiles.js, etc) global so we can access it anywhere.

test.js
```diff
@@ -162,66 +201,26 @@ before(async () => {
...
       // Set amountOfHeadersToSend to 500 to avoid having to inform headers in separated calls
       config.customConfig[`federator.amountOfHeadersToSend`] = 500;
 
-      // federatesToStart: [1, 2, 3]
-      // additionalFederateNodes: [4, 5]
-      // Configure peers
-      let peerIndex = 0;
-      for (let i = 0; i < federatesToStart.length; i++) {
-        if (federatesToStart[i] !== federateToStart) {
-          config.customConfig[`peer.active.${peerIndex}.ip`] = '127.0.0.1';
```
jeremy-then, line 191: Hardcoding all this in the config file itself.

config/node-configs/second-federation/fed3.conf
```diff
@@ -12,7 +9,7 @@ federator {
     signers {
         BTC {
           type = "hsm" #hsm or keyFile (keyfile is deprecated)
-          #path = "/var/lib/jenkins/workspace/Pipeline/utilities/configs/keys/reg5.key" #needed when using keyfile
+          #path = "/var/lib/jenkins/workspace/Pipeline/utilities/configs/keys/reg3.key" #needed when using keyfile
```
julia-zack, line 12: same here

lib/bridge-provider.js
```diff
@@ -1,22 +1,13 @@
...
-
-    if(bridgeCache) {
-        return bridgeCache;
-    }
-
-    bridgeCache = precompiledArrowhead600.bridge.build(rskClient);
-
-    return bridgeCache;
-}
+    return precompiledArrowhead600.bridge.build(rskClient);
```
julia-zack, line 9: should we use arrowhead631?

lib/federation-utils.js
```diff
@@ -23,7 +26,51 @@ var compareFederateKeys = (publicKeysA, publicKeysB) => {
...
+
+const stopPreviousFederators = async (newFederationConfig) => {
+
+  // TODO: right now there is only the genesis federation and a second federation. So the previous is the genesis federation.
+  // When more are added, we need to change this logic to dynamically select the previous federation.
+  // I tried with a `switch` statement ready for future changes but sonar complains about it.
+  const previousFederationId = 'genesis-federation';
+
+  Runners.fedRunners.forEach((fedRunner) => {
+      if(fedRunner.options.federationId === previousFederationId) {
```
julia-zack, line 43:
```suggestion
      if (fedRunner.options.federationId === previousFederationId) {
```

lib/federation-utils.js
```diff
@@ -23,7 +26,51 @@ var compareFederateKeys = (publicKeysA, publicKeysB) => {
...
+  });
+
+  Runners.fedRunners = Runners.fedRunners.filter(fedRunner => fedRunner.options.federationId !== previousFederationId);
+  Runners.hosts.federates = Runners.hosts.federates.filter(fedHostInfo => fedHostInfo.federationId !== previousFederationId);
+  Runners.hosts.federate = Runners.hosts.federates[0];     
+
+};
+
+const startNewFederationNodes = async (newFederationConfig) => {
+  for(let i = 0; i < newFederationConfig.length; i++) {
```
julia-zack, line 55:
```suggestion
  for (let i = 0; i < newFederationConfig.length; i++) {
```

lib/federation-utils.js
```diff
@@ -23,7 +26,51 @@ var compareFederateKeys = (publicKeysA, publicKeysB) => {
...
+      const fedConfig = newFederationConfig[i];
+      await federateStarter.startFederate(i + 1, fedConfig);
+  }
+};
+
+const fundNewFederators = async (rskTxHelper, newFederationConfig) => {
+  for(const fedConfig of newFederationConfig) {
```
julia-zack, line 62:
```suggestion
  for (const newFederatorConfig of newFederationConfig) {
```

lib/federation-utils.js
```diff
@@ -23,7 +26,51 @@ var compareFederateKeys = (publicKeysA, publicKeysB) => {
...
+      const fedRskCompressedPublicKey = fedConfig.publicKeys.rsk;
+      const fedRskAddress = rskUtils.getAddressFromUncompressedPublicKey(rskUtils.uncompressPublicKey(fedRskCompressedPublicKey));
```
julia-zack, line 64:
```suggestion
      const newFederatorRskCompressedPublicKey = fedConfig.publicKeys.rsk;
      const newFederatorRskAddress = rskUtils.getAddressFromUncompressedPublicKey(rskUtils.uncompressPublicKey(fedRskCompressedPublicKey));
```

no big deal, just find it more accurate

lib/rsk-utils.js
```diff
@@ -559,6 +567,33 @@ const waitForRskMempoolToGetAtLeastThisManyTxs = async (rskTxHelper, atLeastExpe
...
+const keccak256 = (str) => {
+  return Web3.utils.keccak256(str);
+}
+
+const removeCompressionPrefix = (uncompressedPublicKey) => {
+  if(uncompressedPublicKey.startsWith('04')) {
+      return uncompressedPublicKey.substring(2);
+  } else if(uncompressedPublicKey.startsWith('0x04')) {
+      return uncompressedPublicKey.substring(4);
+  }
```
julia-zack, line 584:
```suggestion
  if (uncompressedPublicKey.startsWith('04')) {
      return uncompressedPublicKey.substring(2);
  }
  if (uncompressedPublicKey.startsWith('0x04')) {
      return uncompressedPublicKey.substring(4);
  }
```

no need for the `else`, since you are returning right?

julia-zack, line 584: a nit one: 

```suggestion
  if (uncompressedPublicKey.startsWith('0x04')) {
      return uncompressedPublicKey.substring(4);
  }
  if (uncompressedPublicKey.startsWith('04')) {
      return uncompressedPublicKey.substring(2);
  } 
```

i find this way more straightforward

lib/tests/activate-fork
julia-zack: no big deal, but it may be better to first do a pr against the `svp-integration` branch just with this change

lib/bridge-provider.js
jeremy-then: In fact, this will be removed once the precompiled-abis changes are published.
By the way, I think we should use an if here or pass the abi, because if lovell is active, it should use the lovell one and not arrowhead.

config/node-configs/second-federation/fed3.conf
jeremy-then: Updated.

lib/federation-utils.js
jeremy-then: So nit 🤣 .
In fact, that is my style. Do we have a document already for code style?

lib/federation-utils.js
jeremy-then: Same as the other.
In fact, I checked and we are using both ways in the tests. We should agree on what style to use and then refactor them all.

lib/rsk-utils.js
```diff
@@ -1,9 +1,11 @@
 const expect = require('chai').expect;
+const Web3 = require('web3');
```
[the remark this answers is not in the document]
jeremy-then: Not now. Out of the scope. We will need to come back and do a major refactor for the web3 stuff.
The `rsk-transaction-helper` needs to be changed first as well.

lib/rsk-utils.js
jeremy-then: Yes, and now.
With the current logic, what was passed will be returned intact if no condition is matched.
If we remove that `else if`, it will then remove the first 4 characters of the passed string no matter what.

jeremy-then: I agree with this one.

lib/tests/2wp.js
```diff
@@ -472,9 +456,15 @@ const execute = (description, getRskHost) => {
 
         // To ensure no refund is made, let's try to push any possible refund pegout.
         await triggerRelease(rskTxHelpers, btcTxHelper);
+        
+        const isLovellActive = await Runners.common.forks.lovell700.isAlreadyActive();
 
-        await assertPeginTxHashNotProcessed(btcPeginTxHash);
-
+        if(isLovellActive) {
```
julia-zack, line 462:
```suggestion
        if (isLovellActive) {
```

lib/rsk-utils.js
julia-zack: 🤔

lib/tests/2wp.js
```diff
@@ -769,7 +762,9 @@ const execute = (description, getRskHost) => {
 
       // Assert
 
-      await assertExpectedReleaseRequestRejectedEventIsEmitted(senderRecipientInfo.rskRecipientRskAddressInfo.address, pegoutValueInSatoshis, PEGOUT_REJECTION_REASONS.LOW_AMOUNT);
+      const isLovellActive = await Runners.common.forks.lovell700.isAlreadyActive();
```
julia-zack, line 765: can't you set it as a global variable? does that make sense?

lib/federation-utils.js
julia-zack, line 62:
```suggestion
  for(const newFederatorConfig of newFederationConfig) {
```

lib/rsk-utils.js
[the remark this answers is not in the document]
julia-zack: makes sense

lib/tests/2wp.js
julia-zack, line 1: i don't know if its possible, but i think it may be a good idea to have a first pr with the refactors, and then another one with lovell changes, to keep the changes more organized. Wdyt?

jeremy-then: Yeah, but its already done. It would take much more effort now to undo part of this work.
I did it this way because I wanted to leave the tests passing. And it got complicated anyways.

lib/tests/activate-fork
jeremy-then: I would agree but that would complicate and delay things much more.
I wanted to leave the tests in a passing state.

lib/federation-utils.js
jeremy-then: Updated.

lib/bridge-provider.js
nathanieliov: Pass the abi sounds better, no?

lib/tests/activate-fork
```diff
@@ -0,0 +1,29 @@
...
+          const isForkAlreadyActive = await fork.isAlreadyActive();
+          expect(isForkAlreadyActive, `Fork ${fork.name} not yet active`).to.be.true;
+        } catch (err) {
+          throw new CustomError('Block mining until reach height to activate fork failure', err);
+        }
+      });
+  });
+}
+
+module.exports = { execute };
```
nathanieliov, line 29: Missing an empty line at the end o the file for git.

lib/tests/change-federation.js
```diff
@@ -0,0 +1,287 @@
...
+            newFederationBtcPublicKeys = newFederationPublicKeys.map(federator => federator[KEY_TYPE_BTC]);
+
+            const expectedNewFederationErpRedeemScriptBuffer = redeemScriptParser.getP2shErpRedeemScript(newFederationBtcPublicKeys.map(key => removePrefix0x(key)), ERP_PUBKEYS, ERP_CSV_VALUE);
+            expectedNewFederationErpRedeemScript = expectedNewFederationErpRedeemScriptBuffer.toString('hex');
+            expectedNewFederationAddress = redeemScriptParser.getAddressFromRedeemScript('REGTEST', expectedNewFederationErpRedeemScriptBuffer);
+
+            initialFederationAddress = await bridge.methods.getFederationAddress().call();
+
+            await btcTxHelper.importAddress(initialFederationAddress, 'federations');
+            await btcTxHelper.importAddress(expectedNewFederationAddress, 'federations');
```
nathanieliov, line 121: Is there not prob with both having the same label?

nathanieliov (approved):
Well done. Just some minor comments

lib/bridge-provider.js
jeremy-then: Since we are only going to have the latest old fork and the new incoming fork, I will add a check for that better. There's no need to work with previous abis, just a maximum of 2.

config/regtest-all-keyfiles.js
```diff
@@ -0,0 +1,222 @@
+const path = require('path');
+const nodesConfigPath = 'config/node-configs';
+const keysPathResolve = 'node-keys';
+const powpegNodeJarPath = process.env.POWPEG_NODE_JAR_PATH;
+const federatesLogbackPath = process.env.LOG_HOME;
+
+const bookkeepingConfigurations = {
+    difficultyTarget: 3,
+    informerIntervalInMs: 2000,
+    blockHeadersToSend: 27
```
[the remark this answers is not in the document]
jeremy-then: Those are the values as they have always been in this project. I just copied and pasted this from another file.

configure.sh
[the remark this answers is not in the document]
jeremy-then: That's weird. It has that line. It might be complaining about the old code not having that line I think.

lib/tests/change-federation.js
jeremy-then: No, because they are both part of the federation. At least this is how they are defined in the old `04_00_02-fedchange.js` file.

lib/tests/activate-fork
```diff
@@ -0,0 +1,29 @@
+const { expect } = require('chai');
+const rskUtils = require('../rsk-utils');
+const CustomError = require('../CustomError');
+const { getRskTransactionHelpers } = require('../rsk-tx-helper-provider');
+
+const execute = (fork, getFederates) => {
```
marcos-iov, line 6: Maybe out of scope for now, but keep in mind we will only be activating one fork. The upcoming one. So no need to make this too generic

config/regtest-all-keyfiles.js
```diff
@@ -0,0 +1,222 @@
+const path = require('path');
```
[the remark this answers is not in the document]
jeremy-then: Yes, a bit.

lib/tests/activate-fork
jeremy-then: Gotcha!

config/node-configs/second-federation/fed2.conf
```diff
@@ -0,0 +1,69 @@
+
+peer {
+    privateKey = "cb0878d04ae1d7d30df5866f845467b1d3ead5116b7732b894d69b89dc90f03a"
+}
+
+# federator configuration
+federator {
+    enabled = true
+    signers {
+        BTC {
```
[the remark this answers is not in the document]
jeremy-then: I noticed that that config was already in the `regtest-all-keyfiles.js` file. Removed from here.

config/node-configs/second-federation/fed2.conf
```diff
@@ -0,0 +1,69 @@
+
+peer {
+    privateKey = "cb0878d04ae1d7d30df5866f845467b1d3ead5116b7732b894d69b89dc90f03a"
```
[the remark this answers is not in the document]
jeremy-then: I'm not sure now. I don't think so.

config/node-configs/second-federation/fed2.conf
```diff
@@ -0,0 +1,69 @@
...
+       }
+    }
+    bitcoinPeerAddresses = [
+        "127.0.0.1:18444"
+    ]
+    gasPrice = 1000
+
+    updateBridgeBtcBlockchain = true
+    updateBridgeBtcCoinbaseTransactions = true
+    updateBridgeBtcTransactions = true
```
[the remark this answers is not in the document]
jeremy-then: Removed.

lib/bridge-provider.js
nathanieliov: Cool. LGTM

marcos-iov (approved):

PR #271, opened 2025-06-06
Initial reed tests setup

jeremy-then:
`rskj:segwit_compatible_powpeg_integration`
`fed:fix-update-bridge-method`

This pr shows the initial setup to test reed and try to do a segwit federation change.


lib/tests/2wp.js
```diff
@@ -67,7 +67,7 @@ const execute = (description) => {
     after(async () => {
 
         // Ensure there's no pegout txs left to be processed
-        await triggerRelease(rskTxHelpers, btcTxHelper);
+        await triggerRelease(rskTxHelpers, btcTxHelper, {}, false);
```
jeremy-then, line 70: When there's no `release_request_received`, then there's no reason to mine extra blocks for "next pegout creation block number", so passing a `false` flag to avoid mining extra blocks in this case.

marcos-iov (approved):

PR #385, opened 2026-04-09
added a test for the getEstimatedFeesForNextPegoutEvent

julianlen:


lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,167 @@
...
+        await triggerRelease(rskTxHelpers, btcTxHelper);
+
+        const countAfterRelease = await getQueuedPegoutsCount();
+        expect(countAfterRelease).to.equal(0);
+    };
+
+    describe(description, () => {
+        before(async () => {
+            rskTxHelper = getRskTransactionHelper();
+            rskTxHelpers = getRskTransactionHelpers();
```
jeremy-then:
```suggestion
            rskTxHelpers = getRskTransactionHelpers();
            rskTxHelper = rskTxHelpers[0];
```

Like we have been doing here: https://github.com/rsksmart/rootstock-integration-tests/blob/9eb1e5a1143e7a9be3d1bf93f418b8309abb02a1/lib/tests/2wp.js#L65-L66

And remove the `getRskTransactionHelper` import above.

What do you think?

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,167 @@
...
+            btcTxHelper = getBtcClient();
+            bridge = await getBridge(rskTxHelper.getClient());
+        });
+
+        beforeEach(async () => {
+            await triggerRelease(rskTxHelpers, btcTxHelper);
```
jeremy-then: What do you think about leaving a comment above this line, explaining why we call this here.

Something like:

// Ensuring there's no pegout pending in the Bridge.

Or anything more descriptive that you might come up with. Up to you.

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,167 @@
...
+            await setupFedUtxosAndCreateAPegoutRequest();
+        });
+
+        afterEach(async () => {
+            await triggerRelease(rskTxHelpers, btcTxHelper);
```
jeremy-then: I don't think we need to run `triggerRelease` before and after, doesn't make a difference because it would run twice in a row after the first test.

I think with the one `beforeEach` is enough. That would take care of it.

What do you think?

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,167 @@
...
+    let btcTxHelper;
+    let senderInfo;
+
+    const getQueuedPegoutsCount = async () =>
+        Number(await bridge.methods.getQueuedPegoutsCount().call());
+
+    const getEstimatedFeesForNextPegoutEvent = async () =>
+        Number(await bridge.methods.getEstimatedFeesForNextPegOutEvent().call());
+
+    const setupFedUtxosAndCreateAPegoutRequest = async () => {
```
jeremy-then: No big deal maybe, but should we add some documentation explaining a bit the usage of this function?

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,167 @@
...
+        );
+
+        await sendTxToBridge(
+            rskTxHelper,
+            new BN(satoshisToWeis(MINIMUM_PEGOUT_AMOUNT_IN_SATOSHIS)),
+            senderInfo.rskRecipientRskAddressInfo.address
+        );
+    };
+
+    const assertReleaseRequestsAreProcessed = async () => {
```
jeremy-then: I think this function does more than just 'assert'. It actually ensures the pegouts are processed by calling `triggerRelease`.
What do you think of renaming it something like: `ensureReleaseRequestsAreProcessed` or something like that? No big deal maybe, we could leave it like this if not a problem and the team agrees.

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/319267654d7231ae1c2149737d3337f70d4f753d

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/6b657f43ac8f7b13dd2771a0488856f97e2dbe31

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/11118678ebf8098dc7c9f1e4680acbf8cca06012

lib/tests/get_estimated_fees_methods.js
julianlen: processReleaseRequestsAndAssertProcessed?

julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/16a933f9131ca0c55505b4a7421e332b689ee101

lib/tests/get_estimated_fees_methods.js
julianlen: If the name isn't clear enough I'd think about a better name instead of a comment, wdyt?

julianlen: check https://github.com/rsksmart/rootstock-integration-tests/pull/385/changes/0a3543831cbc351191ce94fdddd6e93a1543ff2d

jeremy-then: I was thinking about the technical decision and need of this function. Usually reading the name doesn't tell you the why. I had to read the whole implementation to understand what it was doing, and one needs to have a lot of context to understand the need to have this function.

What I'm suggesting is the latest approach we came up with while refactoring this project.
For example, notice this: https://github.com/rsksmart/rootstock-integration-tests/blob/9eb1e5a1143e7a9be3d1bf93f418b8309abb02a1/lib/rsk-utils.js#L410-L414

And most of the utility functions in rsk-utils.js.

jeremy-then (dismissed):
LGTM.

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,168 @@
...
+    let bridge;
+    let rskTxHelper;
+    let rskTxHelpers;
+    let btcTxHelper;
+    let senderInfo;
+
+    const getQueuedPegoutsCount = async () =>
+        Number(await bridge.methods.getQueuedPegoutsCount().call());
+
+    const getEstimatedFeesForNextPegoutEvent = async () =>
```
julia-zack: super nit:
```suggestion
    const getEstimatedFeesForNextPegOutEvent = async () =>
```

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,168 @@
...
+                }
+            });
+        });
+
+        describe('getEstimatedFeesForPegOutAmount', () => {
+            it('should allow constructing a second pegout request using the estimated fees from getEstimatedFeesForPegOutAmount and process both', async () => {
+                try {
+                    const estimatedFeesForPegout = Number(await bridge.methods
+                        .getEstimatedFeesForPegOutAmount(satoshisToWeis(MINIMUM_PEGOUT_AMOUNT_IN_SATOSHIS))
+                        .call());
```
julia-zack: should we extract the call as we do for the other methods?

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,168 @@
...
+                        'Error validating getEstimatedFeesForNextPegOutEvent with two pegouts',
+                        err
+                    );
+                }
+            });
+        });
+
+        describe('getEstimatedFeesForPegOutAmount', () => {
+            it('should allow constructing a second pegout request using the estimated fees from getEstimatedFeesForPegOutAmount and process both', async () => {
+                try {
```
julia-zack: is the try-catch needed?

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,168 @@
...
+                        .call());
+                    expect(estimatedFeesForPegout).to.be.greaterThan(0);
+
+                    const secondPegoutAmountInSatoshis =
+                        MINIMUM_PEGOUT_AMOUNT_IN_SATOSHIS + estimatedFeesForPegout;
+                    await sendTxToBridge(
+                        rskTxHelper,
+                        new BN(satoshisToWeis(secondPegoutAmountInSatoshis)),
+                        senderInfo.rskRecipientRskAddressInfo.address
+                    );
```
julia-zack: should we extract this? exact same code for both tests

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,168 @@
...
+        );
+
+        await sendTxToBridge(
+            rskTxHelper,
+            new BN(satoshisToWeis(MINIMUM_PEGOUT_AMOUNT_IN_SATOSHIS)),
+            senderInfo.rskRecipientRskAddressInfo.address
+        );
+    };
+
+    const processReleaseRequestsAndAssertQueueCleared = async () => {
```
julia-zack: this method is acting and asserting. Can we separate it?

lib/tests/get_estimated_fees_methods.js
julia-zack, line 1: missing a bunch of test cases: 0 requests, multiple requests, different pegout amounts.

lib/tests/get_estimated_fees_methods.js
julianlen: even if it's used in one place (with the call)?

lib/tests/get_estimated_fees_methods.js
julianlen: this wasn't the idea, it is an integration test

lib/tests/get_estimated_fees_methods.js
jeremy-then: I agree because that's exactly how it is in the Bridge.

lib/tests/get_estimated_fees_methods.js
jeremy-then: I would agree because is being used in 2 places:`'should allow constructing a second pegout request using the estimated fees from getEstimatedFeesForPegOutAmount and process both'` and `'should revert when peg-out amount in weis is below minimum peg-out'`.

Not a big deal though.

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/changes/6335839a3f0d7b7667eddb17dceafcf027e44323

lib/tests/get_estimated_fees_methods.js
jeremy-then: Probably not. That's the old way of doing it. And now that I think about it, I remember one of the reasons we stop adding try/catch.

Each assertion should have a good and descriptive message to make it easy and natural to debug and understand what happened.

Something like this: https://github.com/rsksmart/rootstock-integration-tests/blob/9eb1e5a1143e7a9be3d1bf93f418b8309abb02a1/lib/tests/change-federation.js#L314-L316

So we stop seeing assertion failures like: `expected false to be true` or `expected 1 to be 0` and instead see something like this: `Expected the pegout queue to be empty by now, but it contains 1 pegout`. Etc.

@julianlen , @julia-zack , what do you think?

lib/tests/get_estimated_fees_methods.js
jeremy-then: I agree.

lib/tests/get_estimated_fees_methods.js
julia-zack: so much better that way yes. Using the expect with a descriptive message should be the correct way to do it

julianlen: here is the change https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/748854064f86c6f04cb2e8bd4cf0a8fce4843663

Even when I had the try/catch, when it failed because it hadn't found the getEstimatedFeesForPegOutAmount method, the error wasn't clear

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/3b7148f777db08431270470c2a033d1f3ad11bd1

lib/tests/get_estimated_fees_methods.js
jeremy-then: In the middle of 2 assertions it has a `triggerRelease` call, which is the 'act' part you are referring to. I don't see an issue with this function. Separating it wouldn't probably improve much. We could probably rename it. There are many utility functions doing assertions and processing stuff. That's the nature of this project actually. It's a testing library. For example, `sendFromCow` does stuff and also asserts.
`expect` is used throughout this project to simply check stuff as well. Check also `startFederate`, `disableWhitelisting`.

lib/tests/get_estimated_fees_methods.js
julia-zack: you should add the error message you wanna see in the `expect` section, as jere said above

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/d8a1d241bd8d027ae19112ff3013a74d7b70cc2f

lib/tests/get_estimated_fees_methods.js
julia-zack: i still dont love it, but feel free to leave it as is.

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,148 @@
...
+        });
+
+        // Ensuring there's no pegout pending in the Bridge contract
+        // after each test to avoid interference between tests
+        afterEach(async () => {
+            await triggerRelease(rskTxHelpers, btcTxHelper);
+        });
+
+        describe('getEstimatedFeesForNextPegOutEvent', () => {
+            it('should allow constructing a second pegout request using the fees from getEstimatedFeesForNextPegOutEvent and process both', async () => {
```
julia-zack:
```suggestion
            it('should allow constructing a second pegout request using the estimated fees from getEstimatedFeesForNextPegOutEvent and process both', async () => {
```
nit, but to match other test name

lib/tests/get_estimated_fees_methods.js
julianlen: wdyt something like this

```
await assertQueuedPegoutsCount(2);
await releaseQueuedPegouts();
await assertQueuedPegoutsCount(0);
```

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/1f2080e0c220b908988103aa6b9d76887c339cb7

lib/tests/get_estimated_fees_methods.js
julia-zack: no worries, leave it as is :)

lib/tests/get_estimated_fees_methods.js
julianlen: https://github.com/rsksmart/rootstock-integration-tests/pull/385/changes/98fd3d23ca4af2eccd6a6f7df1b36aaa7efa98e6

added a second part of each test as discussed checking getEstimatedFees with no pegout requests

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,161 @@
...
+        });
+
+        // Ensuring there's no pegout pending in the Bridge contract
+        // after each test to avoid interference between tests
+        afterEach(async () => {
+            await triggerRelease(rskTxHelpers, btcTxHelper);
+        });
+
+        describe('getEstimatedFeesForNextPegOutEvent', () => {
+            it('should build pegouts from getEstimatedFeesForNextPegOutEvent when one pegout is queued and when none are queued, and process all pegouts', async () => {
```
julia-zack: why did you add the 0 request case inside the other ones? thought it would be a separated test case

julia-zack (dismissed):

tests/01_06_51-get_estimated_fees_methods.js
```diff
@@ -0,0 +1,3 @@
+const getEstimatedFeesMethodsTests = require('../lib/tests/get_estimated_fees_methods');
```
marcos-iov, line 1: By the name of the file it will run after activateLatestFork, probably not a good idea.

It would be better to have this along with all the other "normal"
 test scenarios. Tests that go after the activation of the latest fork are for testing the upcoming yet on development network upgrade

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,180 @@
...
+ * @param {string} [description]
+ */
+const execute = (description) => {
+    let bridge;
+    let rskTxHelper;
+    let rskTxHelpers;
+    let btcTxHelper;
+    let senderInfo;
+
+    const shouldReturnGreaterThanZeroErrorMessage = (methodName) =>
```
marcos-iov: Looks like this is only used with the same param always `getEstimatedFeesForNextPegOutEvent` so it probably doesn't make sense to keep it as method, could simply be a constant

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,180 @@
...
+
+        const bridgeStateAfterPegin = await getBridgeState(rskTxHelper.getClient());
+        const peginUtxoInBridgeState = bridgeStateAfterPegin.activeFederationUtxos.find(
+            (utxo) =>
+                utxo.btcTxHash === peginBtcTxHash &&
+                utxo.valueInSatoshis === FEDERATION_PEGIN_AMOUNT_IN_SATOSHIS
+        );
+        expect(peginUtxoInBridgeState).to.not.be.undefined;
+
+        await sendFromCow(
```
marcos-iov: Do we need to send funds from the cow address? Weren't those funds already received from the peg-in?

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,180 @@
...
+            rskTxHelper,
+            senderInfo.rskRecipientRskAddressInfo.address,
+            Number(btcToWeis(PEGOUT_SENDER_FUNDING_IN_BTC))
+        );
+
+        await sendTxToBridge(
```
marcos-iov: Would it make sense to get the bridge state after this and assert there is a pegout request?

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,180 @@
...
+
+        await triggerRelease(rskTxHelpers, btcTxHelper);
+
+        const countAfterRelease = await getQueuedPegoutsCount();
+        expect(countAfterRelease).to.equal(0,
+            'All pegout requests should have been processed and the queue should be cleared after triggering release'
+        );
+    };
+
+    const sendPegoutWithEstimatedFees = async (estimatedFees) => {
```
marcos-iov:
```suggestion
    const sendPegoutWithEstimatedFees = async (estimatedFeesInSatoshis) => {
```
Adding the unit helps, it can get confusing vey fast

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,180 @@
...
+        });
+
+        beforeEach(async () => {
+            await setupFedUtxosAndAddAPegoutRequestToTheBridge();
+        });
+
+        // Ensuring there's no pegout pending in the Bridge contract
+        // after each test to avoid interference between tests
+        afterEach(async () => {
+            await triggerRelease(rskTxHelpers, btcTxHelper);
```
marcos-iov: Get the bridge state and assert there are no pending pegout requests after this?

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,180 @@
...
+        });
+
+        describe('getEstimatedFeesForNextPegOutEvent', () => {
+            it('should build pegouts from getEstimatedFeesForNextPegOutEvent when one pegout is queued and when none are queued, and process all pegouts', async () => {
```
marcos-iov: Not sure I follow this test.

get estimated fees with 1 request and assert it's > 0. Then add a second request and immediately process the pegouts, why?

Then get estimated fees with 0 requests and assert it's > 0.

Finally send a new pegout request and immediately process it.

It all seems a bit confusing

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,180 @@
...
+                    shouldReturnGreaterThanZeroErrorMessage('getEstimatedFeesForNextPegOutEvent')
+                );
+
+                await sendPegoutWithEstimatedFees(estimatedFeesForNextPegOutWithNoEnqueuedRequests);
+                await processReleaseRequestsAndAssertQueueCleared(ONE_PEGOUT_REQUEST_EXPECTED);
+            });
+        });
+
+        describe('getEstimatedFeesForPegOutAmount', () => {
+            it('should build pegouts from getEstimatedFeesForPegOutAmount when one pegout is queued and when none are queued, and process all pegouts', async () => {
```
marcos-iov: Same here.

Also, we should test passsing different values as the param of `getEstimatedFeesForPegOutAmount`. Perhaps we should have 2 utxos in the Bridge? So for one param it only needs 1 utxo and returns a certain value, for another param it needs 2 utxos to build the transaction so the fees are higher

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/commit/78554849aa1f379256387b91eb66e5b913bc423b (it was used with both methods but for some reason I forgot to changed it to getEstimatedFeesForPegOutAmount)

tests/01_06_51-get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/15cf620de5df81c0113863ab35d60c271f508ebc

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/44cefb75e16d300c597df8f818bf9895ab3aa328

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/9bf91f03d002ca728a108f7038924f9e871aab44

lib/tests/get_estimated_fees_methods.js
julianlen: done https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/ea8b9bd5c36cf12de8509889f001d39fe5b1d93a

lib/tests/get_estimated_fees_methods.js
julianlen: finally deleted in https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/8e50c606cbc2c127cb8e794e79b29bdc41f6ab2d

lib/tests/get_estimated_fees_methods.js
julianlen: I deleted this in https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/8e50c606cbc2c127cb8e794e79b29bdc41f6ab2d

lib/tests/get_estimated_fees_methods.js
julianlen: As we discussed, we are changing only the number of pegout requests.

done in https://github.com/rsksmart/rootstock-integration-tests/pull/385/commits/8e50c606cbc2c127cb8e794e79b29bdc41f6ab2d

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * Bitcoin miner fee (satoshis) for the pegout tx in the last `release_btc` event.
+ * Uses matching `pegout_transaction_created.utxoOutpointValues` for input amounts.
+ *
+ * @param {object[]} pegoutsEvents from `getPegoutEventsInBlockRange`
+ * @returns {number} fee in satoshis
+ */
+const getReleaseBtcTransactionFeeInSatoshis = (pegoutsEvents) => {
+    // Get the latest event which is the `release_btc` event
+    const releaseBtcEvents = pegoutsEvents.filter(
```
jeremy-then, line 32: Do we know exactly in what order these events will be? Because if so, then we could do something like this to avoid looping:

https://github.com/rsksmart/rootstock-integration-tests/blob/6e2c505f9cc153eeadb35a068c7371f2823278c2/lib/tests/2wp.js#L1714

https://github.com/rsksmart/rootstock-integration-tests/blob/6e2c505f9cc153eeadb35a068c7371f2823278c2/lib/tests/2wp.js#L1781

https://github.com/rsksmart/rootstock-integration-tests/blob/6e2c505f9cc153eeadb35a068c7371f2823278c2/lib/tests/2wp.js#L1789

https://github.com/rsksmart/rootstock-integration-tests/blob/6e2c505f9cc153eeadb35a068c7371f2823278c2/lib/tests/2wp.js#L1797

https://github.com/rsksmart/rootstock-integration-tests/blob/6e2c505f9cc153eeadb35a068c7371f2823278c2/lib/tests/2wp.js#L1805

https://github.com/rsksmart/rootstock-integration-tests/blob/6e2c505f9cc153eeadb35a068c7371f2823278c2/lib/tests/2wp.js#L1813

https://github.com/rsksmart/rootstock-integration-tests/blob/6e2c505f9cc153eeadb35a068c7371f2823278c2/lib/tests/2wp.js#L1817

If not, then no big deal. My suggestion is just for performance and to ensure the events are in the right order.

lib/tests/get_estimated_fees_methods.js
```diff
@@ -0,0 +1,234 @@
...
+    const btcTx = bitcoinJsLib.Transaction.fromHex(
+        removePrefix0x(latestReleaseBtcEvent.arguments.btcRawTransaction)
+    );
+    const btcTxId = btcTx.getId();
+
+    // Gets the `pegout_transaction_created` event for the pegout tx in the `release_btc` event
+    // to obtain the input amounts of the pegout tx and calculate the fee.
+    const pegoutTxCreated = pegoutsEvents.find(
+        (event) =>
+            event.signature === PEGOUT_EVENTS.PEGOUT_TRANSACTION_CREATED.signature &&
```
jeremy-then, line 47: Same optional suggestion as https://github.com/rsksmart/rootstock-integration-tests/pull/385/changes#r3157872701, if possible.

test.js
```diff
@@ -101,6 +101,7 @@ global.Runners = {
             arrowhead631: createForkObject('arrowhead631', 1),
             lovell700: createForkObject('lovell700', 1),
             reed800: createForkObject('reed800', 1),
+            vetiver900: createForkObject('vetiver900', 1),
```
jeremy-then, line 104: I thought we were going to set a future activation block height so the tests would go as usual, then we use `tests/01_07_00-activate-latest-fork.js` to activate the fork, then run the `tests/01_07_00-activate-latest-fork.js` test. Isn't it?

But right now vetiver900 is active from the beginning.

Let me know if this is intended.

julianlen: https://github.com/rsksmart/rootstock-integration-tests/pull/385#discussion_r3095479120 it is!

lib/tests/get_estimated_fees_methods.js
julianlen: I think this is better so the test is sure we are using a pegoutTransactionCreated event

lib/tests/get_estimated_fees_methods.js
julianlen: couldn't exist the chance that the event doesn't exist or the latest is a different one?

jeremy-then (dismissed):
LGTM.

marcos-iov (approved):

PR #407, opened 2026-04-29
Adds super and base event tests.

jeremy-then:
Adds super and base event tests.

### DIFF
--- a/lib/tests/union-bridge-methods.js
+++ b/lib/tests/union-bridge-methods.js
@@ -326,6 +326,17 @@ const execute = (description) => {
                 const stored = await bridgeMethods.getSuperEvent().call();
                 expect(stored.toLowerCase()).to.equal(payload.toLowerCase());
             });
+
+            it('should accept setBaseEvent at exactly the maximum payload length', async () => {
+                const payload = maxLengthPayloadHex();
+                await rskUtils.sendTransaction(
+                    rskTxHelper,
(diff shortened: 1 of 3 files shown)


test.js
```diff
@@ -101,6 +101,7 @@ global.Runners = {
             arrowhead631: createForkObject('arrowhead631', 1),
             lovell700: createForkObject('lovell700', 1),
             reed800: createForkObject('reed800', 1),
+            reed810: createForkObject('reed810', 1),
```
jeremy-then, line 104: We were missing this fork here, so I added it.

lib/tests/union-bridge-methods.js
```diff
@@ -137,6 +135,199 @@ const execute = (description) => {
...
+            });
+
+            it('should reject setBaseEvent when payload length is above the maximum', async () => {
+                const method = unionBridgeContract.methods.setBaseEvent(tooLongPayloadHex());
+                await rskUtils.assertContractCallFails(method, {
+                    from: unionBridgeContractOwnerAddress,
+                });
+            });
+
+            it('should accept setSuperEvent at exactly the maximum payload length', async () => {
```
marcos-iov, line 319: Missing this case for base event?

jeremy-then: Added here: 8781541251c3e5259880793af90069d00b07052f

marcos-iov (approved):

PR #433, opened 2026-06-22
Fix issue while CI tests are failing to start because of update in RSKJ verification-metadata.xml

wlaclaw:
If there are updates in `verification-metadata.xml` in `rskj` repository it would cause test to fail in any case.

The core reason rskj node will fail to build because build environment has it's own `verification-metadata.xml` that does not respect what `rskj` node has.

This PR added script that is merged new entries from `rskj` `verification-metadata.xml` into custom build environment.


fmacleal (dismissed):
LGTM

italo-sampaio (commented):
Was intriged to find out if this whole xml manipulation was really needed, and apparently we don't need all that.

Is there a good reason why we can't just use [this one liner](https://github.com/rsksmart/rootstock-integration-tests/compare/exp/rit-lenient?expand=1) instead?

I tested [pointing your PR to that branch](https://github.com/rsksmart/rskj/compare/rskip559-peg...exp/rit-lenient-3539?expand=1) and turns out [the RIT run](https://github.com/rsksmart/rskj/actions/runs/28264027901/job/83746288127) (still fail, but that's an actual test failing, not an early abort). This might be enough to fix the issue, and more human-readeable than the xml parsing version.

italo-sampaio (approved):
LGTM! Thanks for taking care of that

marcos-iov (approved):

PR #448, opened 2026-07-10
Ensure SVP spend tx is registered

marcos-iov:
This pull request refactors the logic for waiting on RSK mempool transactions and improves the reliability of the federation change test. The main changes include making the wait functions more flexible by adding an optional callback, and updating the test to use a more robust polling loop for SVP spend transaction registration instead of fixed waits.

**Enhancements to wait functions:**

* Added an optional `onEachAttempt` callback parameter to `waitForRskMempoolToGetThisCountOfThisTxType` and `waitForRskMempoolToGetThisCountOfRegisterBtcTx`, allowing custom logic to be executed on each polling attempt. [[1]](diffhunk://#diff-5ae6cb268232abc556725fd7b7c48ea39c206a51ca4ca4c3b4e122fbe24c741aL250-R251) [[2]](diffhunk://#diff-5ae6cb268232abc556725fd7b7c48ea39c206a51ca4ca4c3b4e122fbe24c741aL282-R297)
* Updated the polling logic to call `onEachAttempt` only when the expected transactions are not yet in the mempool, preventing unnecessary extra submissions.

**Improvements to federation change test reliability:**

* Replaced the previous fixed-wait approach with a polling loop in `lib/tests/change-federation.js` that repeatedly calls `waitAndUpdateBridge` and checks for SVP spend transaction registration, making the test more robust to timing variations.
* Removed redundant code that checked SVP spend transaction registration after the loop, since it is now handled within the polling loop.

### DIFF
--- a/lib/rsk-utils.js
+++ b/lib/rsk-utils.js
@@ -251,9 +251,6 @@ const waitForRskMempoolToGetThisCountOfThisTxType = async (
     onEachAttempt = null
 ) => {
     const method = async () => {
-        if (onEachAttempt) {
-            await onEachAttempt();
-        }
         const mempoolTxs = await getRskMempoolTransactionsToTheBridge(rskTxHelper, true);
         const txsToBridge = mempoolTxs.filter((tx) => tx.input?.startsWith(txType.methodSelector));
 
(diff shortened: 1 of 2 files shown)


jeremy-then (approved):
LGTM.

